PR #86, opened 2021-11-18
Add extensions and docs

anngvu:
This update **DOES NOT** add any new terms directly to `NF.csv`. In `NF.csv`, the only changes are additional columns prefixed with `.` to create a more formal reference for use in visualization and other internal functions. The `NF.jsonld` remains exactly the same, which is why it is not a changed file. New terms and semantics are **not** installed into `NF.csv` directly but as additional files `ext_classes.csv` and  `ext_relations.csv`. The extensions allow better organization/visualization of the schema but doesn't affect annotation, since the terms available are still only in `NF.csv`. 

A new `docs` directory contain R source and relevant output `index.html`, which is intended to be the public docs on gh-pages.


allaway (approved):
Looking good! Approved. Thank you!

PR #90, opened 2021-12-08
Update terms and template docs

anngvu:
This was going to focus on 5-6 but pulled in some term updates along the way.
1. Pruned more deprecated terms that somehow survived
2. Filled in/revised descriptions for some terms
3. Added new `modelSystemName` terms
4. Made `dataType` required -- for some reason it wasn't required and it probably should be
5. Internal organization and notes (for docs)
6. Major additions to the documentation for templates


allaway (approved):
@anngvu , this looks great! I mainly looked at the added values for the purposes of this review.

Would you be willing to walk through the remainder of the changes in our next? I have a general sense of what you are doing with the additions to the schema but a walk through would probably help me understand more thoroughly. 

I am guessing this update would be a minor version in semver world. It's a little hard to translate semver to this type of project but it seems about right?

PR #91, opened 2021-12-08
patch sc93.1 display name

allaway:
I was about to make a comment in https://sagebionetworks.jira.com/jira/servicedesk/projects/NFOSI/queues/custom/7/NFOSI-25 that the term should be available, but then I noticed the slight capitalization difference in the display name, which I missed in my review of #90, my bad.

If this change looks good to you I'll push it to the data curator app today!


anngvu:
Good catch! I need to turn annoying autocaps off or actually use a code editor.

PR #98, opened 2021-12-29
add cellType list validation rule

allaway:
In some scenarios is needed when modelSystemName has multiple values.


allaway:
Tested on staging and appears to work.

PR #100, opened 2022-01-21
Clinical template

anngvu:
Adding this template for annotation of e.g. bloodwork data and simple endpoints from clinical assays that usually come in a table. However, there is currently an issue that's unclear how to work around. Looking at the [spreadsheet test template](https://docs.google.com/spreadsheets/d/1aKUhsSb28J2SmEZZofS5WCjIHEGufaL9YpXXeG6hoxQ/edit#gid=0 ), note how there are a couple of columns after `comments` column that I don't think should be there. These are pulled in automatically because of dependencies defined previously. For example, `modelSystemName` shouldn't be in this template at all because we should always have human data. It appears because `modelSystemName` is in DependsOn for e.g. "rat", which is a valid value for species (see below `schema.json` snippet).

```
{
      "if": {
        "properties": {
          "species": {
            "enum": [
              "Rattus norvegicus"
            ]
          }
        },
        "required": [
          "species"
        ]
      },
      "then": {
        "properties": {
          "modelSystemName": {

```
Normally, this is good, because whenever we reference rat, either through species or something else, bringing in `modelSystemName` can gather additional information. In this case, anything non-human should not be a valid potential value for species, but there isn't a way to get this logic and prevent `modelSystemName` from its eager and unwelcome appearance. Maybe can only ignore it for now?


allaway:
I think it would be reasonable to just ignore it for now, but agree that it's not optimal. I think we could probably fix this by encoding Components properly? https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/2473623559/The+Data+Model+Schema We probably should have done this from the outset, but we might have started this prior to components existing, or at least the docs describing them existing. I cannot recall exactly...

Either way, I think for now - it's totally fine to just ignore. I will merge this and then test it out on the Synodos LGG data (when I get back in the office probably), which has a clinical table that needs annotation!

allaway:
v2.5.0 with these changes is now available in the production app!

PR #103, opened 2022-02-03
add oncotree terms

allaway:
Closes #102, which is a non backwards compatible changes. 

Adds the following terms from ONCOTREE for the Synodos LGG project: 
Pilocytic Astrocytoma
Low Grade Glioma, NOS
Pilomyxoid Astrocytoma
Diffuse Astrocytoma
Glioblastoma Multiforme
Anaplastic Pilocytic Astrocytoma
Pleomorphic Xanthoastrocytoma
Oligoastrocytoma
High-Grade Glioma, NOS
Unknown

Adds the following libraryPreparationMethod:
Illumina TruSeq DNA Nano

Since I was working on tumorTypes already, I also tossed out a bunch of tumor values (mainly blood cancers) that are not currently used in any Portal files. This should help if we decide to import ONCOTREE wholesale down the road, because many of these would need to be changed. 

When available, I converted the remaining terms to ONCOTREE values. The terms didn't actually need to change, just needed to update description to ONCOTREE abbreviation and change source. 

Since I was already making a non b/w compatible change with #102, I also changed `teratoma`, to the ONCOTREE `Teratoma`. 

After this is merged, can remap all instances of Low Grade Glioma to `Low Grade Glioma, NOS` and teratoma to `Teratoma`.

PR #104, opened 2022-02-04
Fix oncotree bug

allaway:
Whoops, I forgot to add the new tumor types in #103 as valid values... this fixes that.


allaway:
Not sure why the diff is so funky, but I'm going to start from a clean slate.

PR #105, opened 2022-02-04
Fix oncotree bug

allaway:
Whoops, I forgot to add the new tumor types in https://github.com/nf-osi/nf-metadata-dictionary/pull/103 as valid values... this fixes that.


anngvu:
Hey, I implemented a different patch while I was adding "age". I noticed that the comma caused a problem, as you can see:

```
{
                    "@id": "bts:LowGradeGlioma"
                },
                {
                    "@id": "bts:NOS"
```
So I think what we'll have to do is use a label without a comma.

allaway:
Oh, that's interesting. I am fairly sure we can have the comma if we wrap the value in quotes under `Valid Values`. If you are working on the same fix, I'll just close and wait for your PR. :)

NF.jsonld
```diff
@@ -3549,26 +3552,50 @@
...
-                    "@id": "bts:Meningioma"
+                    "@id": "bts:Oligoastrocytoma"
                 },
                 {
-                    "@id": "bts:Normal"
+                    "@id": "bts:High-GradeGlioma"
                 },
                 {
-                    "@id": "bts:Teratoma"
+                    "@id": "bts:NOS"
```
anngvu, line 3583: Quoting doesn't work

allaway: Interesting! I could have sworn that works, but obviously mistaken. Thanks for checking!

PR #106, opened 2022-02-04
Patch PR#103, add age-related terms

anngvu:
Noticed valid values also, but had to update the NOS-type tumors as explained [here](https://github.com/nf-osi/nf-metadata-dictionary/pull/105#issuecomment-1029544609).


allaway (approved):
Looks good to me. Thank you!

PR #108, opened 2022-02-04
oncotree patch

allaway:
Sorry for all the back and forth on this. I went to test this in staging and found a typo in LGGNOS that I made on my first PR, and that Unknown was missing from the valid values. This PR corrects those minor differences.

This is not urgent, I'm guessing you're likely done for the day :) 

One random question - any idea why I am getting such a huge diff in the csv after modifying only two lines? I couldn't spot differences between the deleted and added lines, but maybe I'm just missing something. Or maybe diff support for csvs isn't good, but I recall it being better than this...


anngvu:
I think the csv diff thinks the whole file changed because of a missing newline char (similar [stackoverflow](https://stackoverflow.com/questions/1889559/git-diff-to-ignore-m)). But changes in jsonld look fine.

allaway:
Ahh, makes sense. I missed that. Thanks!

PR #110, opened 2022-02-11
add more LGG tumors

allaway:
I am working on updating the Columbia LGG metadata to bring it up to spec. We have detailed tumor type names in that project (annotated as "histology"). I am going to update the tumorType metadata on these files to have these more accurate names, and to make it easier to compare to the Synodos LGG project tumors. However, I needed to add these terms to the schema.


allaway:
Also caught a small bug- `Low Grade Glioma` was still a valid value.

anngvu (approved):

PR #111, opened 2022-02-24
Add new tumor class Subcutaneous Neurofibroma

anngvu:


NF.jsonld
```diff
@@ -14746,74 +14760,6 @@
             "sms:required": "sms:false",
             "sms:validationRules": []
         },
-        {
-            "@id": "bts:AnaplasticPleomorphicXanthoastrocytoma",
```
allaway, line 14750: I may be reading the diff wrong but it looks like this block of tumor types was removed in this version? (I couldn't find it in the CSV either)

allaway (changes requested):
I think some tumors may have gotten removed in this version, can you take a look? I am possibly just reading the diff incorrrectly.

NF.jsonld
anngvu: The diffs can be really weird. If you go to the actual files at that commit, those tumors are untouched: [csv](https://github.com/nf-osi/nf-metadata-dictionary/blob/c8dbe5c0752ff5600decbd83e8cce5e815733080/NF.csv), [jsonld](https://github.com/nf-osi/nf-metadata-dictionary/blob/c8dbe5c0752ff5600decbd83e8cce5e815733080/NF.jsonld)

allaway (changes requested):

NF.jsonld
allaway: Oops, I meant to directly reply to this rather than re-reviewing!

anngvu: Whoops, you're right -- I was confirming Anaplastic Pilocytic Astrocytoma. Figured out that I was checking out files from an outdated branch.

anngvu:
OK, diff should make more sense now. Sorry about this being less efficient than it should have been -- the cognitive cost of switching between 10 terminal windows.

allaway (approved):
LGTM!

allaway:
> OK, diff should make more sense now. Sorry about this being less efficient than it should have been -- the cognitive cost of switching between 10 terminal windows.

No problem, been there myself :)

PR #113, opened 2022-02-25
Create schematic-schema-convert.yml

allaway:
@anngvu, this PR adds a job to automatically build the jsonld file when changes are pushed to the csv. Let me know what you think/if this seems like a reasonable SOP going forward!

### DIFF
--- a/.github/workflows/schematic-schema-convert.yml
+++ b/.github/workflows/schematic-schema-convert.yml
@@ -1,0 +1,69 @@
+name: schematic-schema-convert
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches:    
+      - '**'        # matches every branch
+      - '!main'   # excludes main
+
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


allaway:
BTW, the most recent commit failed because there was no change to the csv (and thus no updated jsonld to commit). https://github.com/nf-osi/nf-metadata-dictionary/runs/5339259981?check_suite_focus=true is an example of a successful run.

anngvu:
Suggestion: There may be cases where we're just updating LICENSE, README, docs. Maybe you can trigger this workflow only if `NF.csv` changes (instead of running it on no changes and getting an error). See official doc for the file/path filter [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths). Also [this example](https://github.com/nf-osi/jobs/blob/main/.github/workflows/docker-update.yml) of using the path filter to make workflow more specific.

allaway:
@anngvu , great idea! Done.

anngvu:
Awesome!

PR #121, opened 2022-03-28
Updates to tumorType

anngvu:


NF.csv
```diff
@@ -601,6 +600,17 @@ Ganglioglioma,GNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
 Anaplastic Ganglioglioma,AGNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
 Optic Pathway Glioma,A glioma that affects the optic nerve. This condition can be seen in association with neurofibromatosis 1. It is most commonly seen in the pediatric age group. ,,,FALSE,https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C4537,tumorType,,,,,,,,,,
 Not Applicable,,,,FALSE,Sage Bionetworks,tumorType,,,,,,,,,,
+Colorectal Carcinoma,A malignant epithelial neoplasm that arrises from the colon or rectum and invates through the muscularis mucosa into the submucosa.,,,FALSE,obo:NCIT_C2955,tumorType,,,,,,,,,,
+Recurrent MPNST,The reemergence of malignant peripheral nerve sheath tumor after a period of remission.,,,FALSE,obo:NCIT_C8823,tumorType,,,,,,,,,, 
+Juvenile Myelomonocytic Leukemia,A myelodysplastic/myeloproliferative neoplasm of childhood that is characterized by proliferation principally of the granulocytic and monocytic lineages,,,FALSE,obo:NCIT_C9233,tumorType,,,,,,,,,,
```
allaway, line 605: JMML is captured in ONCOTREE, could we use that instead? 

Also, it looks like oncotree is missing the NCIT mapping for JMML...I will raise an issue

NF.csv
```diff
@@ -601,6 +600,17 @@ Ganglioglioma,GNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
...
+Sarcoma,"An usually aggressive malignant neoplasm of the soft tissue or bone...Representative subtypes are liposarcoma, leiomyosarcoma, osteosarcoma, and chondrosarcoma.",,,FALSE,obo:NCIT_C9118,tumorType,,,,,,,,,,
+Fibromatosis,A poorly circumscribed neoplasm arising from the soft tissues. It is characterized by the presence of spindle-shaped fibroblasts and an infiltrative growth pattern.,,,FALSE,obo:NCIT_C3042,tumorType,,,,,,,,,,
+Fibrosarcoma,A malignant mesenchymal fibroblastic neoplasm affecting the soft tissue and bone.,,,FALSE,obo:NCIT_C3043,tumorType,,,,,,,,,,
```
allaway, line 608: Can we use ONCOTREE, FIBS?

NF.csv
```diff
@@ -601,6 +600,17 @@ Ganglioglioma,GNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
...
+NF1-Associated Tumor,"Unspecified NF1-associated tumor; use a more specific tumor classification if available.",,,FALSE,Sage Bionetworks,tumorType,,,,,,,,,,This was added to accomodate records that only specified 'NF1' and is patterned after the existence of the NCIT term 'NF1-Associated Malignant Peripheral Nerve Sheath Tumor'
+NF2-Associated Tumor,"Unspecified NF2-associated tumor; use a more specific tumor classification if available.",,,FALSE,Sage Bionetworks,tumorType,,,,,,,,,,This was added to accomodate records that only specified 'NF2' and is patterned after the existence of the NCIT term 'NF1-Associated Malignant Peripheral Nerve Sheath Tumor'
+ANNUBP,"Atypical neurofibromatous neoplasms of uncertain biologic potential is a provisional classification for lesions displaying at least two features: atypia, loss of neurofibroma architecture, high cellularity, and/or mitotic activity >1/50 but <3/10 high power fields. They can be inconsistently diagnosed as atypical neurofibroma or low-grade MPNST. See https://pubmed.ncbi.nlm.nih.gov/28551330/.",,,FALSE,Sage Bionetworks,tumorType,,,,,,,,,,
```
allaway, line 613: We should definitely include ANNUBP. I am wondering what the semantic overlap between ANNUBP and Atypical Neurofibroma is. Can we treat these as the same? IIRC the Atypical Neurofibroma term comes from JHU. If possible, I'd like to merge these two terms.

allaway (changes requested):
In general this is great! I have a few requested changes, and one comment about merging Atypical Neurofibroma and ANNUBP, if possible.

NF.csv
anngvu: In the paper the authors want to distinguish it from Atypical Neurofibroma, which is why the term is added. Sounds it should be a subclass.

NF.csv
```diff
@@ -601,6 +600,17 @@ Ganglioglioma,GNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
 Anaplastic Ganglioglioma,AGNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
 Optic Pathway Glioma,A glioma that affects the optic nerve. This condition can be seen in association with neurofibromatosis 1. It is most commonly seen in the pediatric age group. ,,,FALSE,https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C4537,tumorType,,,,,,,,,,
 Not Applicable,,,,FALSE,Sage Bionetworks,tumorType,,,,,,,,,,
+Colorectal Carcinoma,A malignant epithelial neoplasm that arrises from the colon or rectum and invates through the muscularis mucosa into the submucosa.,,,FALSE,obo:NCIT_C2955,tumorType,,,,,,,,,,
+Recurrent MPNST,The reemergence of malignant peripheral nerve sheath tumor after a period of remission.,,,FALSE,obo:NCIT_C8823,tumorType,,,,,,,,,,
```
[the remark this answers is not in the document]
anngvu: It's defined in NCIT and that was the annotation, so I think adding it is the best way to handle. And it would avoid adding `isRecurrent`.

NF.csv
anngvu: I'll add it to the description.

NF.csv
allaway: NCIT is a great resource and I definitely default to it for most biomedical data, but (IMO) ONCOTREE has a much more well defined tumor hierarchy and semantics, whereas NCIT has a lot of tumor terms that are overlapping. From a practical perspective, ONCOTREE will also be required to upload data to cBioPortal. 

At the end of the day, I think it's fine to go with one or the other, but if we switch to NCIT for these terms we should probably change the other ONCOTREE terms in our dictionary to NCIT, and we should also be sure to use NCIT calues that can be translated to ONCOTREE (https://github.com/cBioPortal/oncotree/blob/master/scripts/ontology_to_ontology_mapping_tool/ontology_mappings.txt)

NF.csv
```diff
@@ -601,6 +600,17 @@ Ganglioglioma,GNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
 Anaplastic Ganglioglioma,AGNG,,,FALSE,ONCOTREE,tumorType,,,,,,,,,,
 Optic Pathway Glioma,A glioma that affects the optic nerve. This condition can be seen in association with neurofibromatosis 1. It is most commonly seen in the pediatric age group. ,,,FALSE,https://ncit.nci.nih.gov/ncitbrowser/ConceptReport.jsp?dictionary=NCI_Thesaurus&ns=ncit&code=C4537,tumorType,,,,,,,,,,
 Not Applicable,,,,FALSE,Sage Bionetworks,tumorType,,,,,,,,,,
+Colorectal Carcinoma,A malignant epithelial neoplasm that arrises from the colon or rectum and invates through the muscularis mucosa into the submucosa.,,,FALSE,obo:NCIT_C2955,tumorType,,,,,,,,,,
```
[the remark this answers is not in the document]
anngvu: I did add Colorectal Adenocarcinoma for future use. Yeah, wish the current annotation was more specific (it's 95% likely Colorectal Adenocarcinoma), but the annotation is what it is.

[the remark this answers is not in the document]
allaway: Such is life. :)

allaway:
Is it just me, or is the schema conversion job not running as expected?

allaway:
Oh, nevermind, I was looking at the wrong thing.

allaway (approved):

Issue #164, opened 2022-06-06
Add [10x Visium Spatial Gene Expression] to [platform]


anngvu:
**List the new value**
Value: `10x Visium Spatial Gene Expression`

The only ontology source is: 
https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0010961

Also vendor source as reference:
https://www.10xgenomics.com/products/spatial-gene-expression

**Additional comments**
This is related to and should have been resolved for issue https://github.com/nf-osi/nf-metadata-dictionary/issues/109

We just added generic "spatial transcriptomics" to assay, so suggest adding this as the specific platform for running the assay (under EFO, it is both assay and a library prep).

PR #165, opened 2022-06-06
Update assay module and document

anngvu:
Mainly issue https://github.com/nf-osi/nf-metadata-dictionary/issues/164 and updating docs to surface new terms.


allaway (approved):
Looks good to me!

PR #224, opened 2023-01-05
Patch/NFOSI-381

anngvu:
Based on a manifest from referenced issue, adds platform, cell line, and proposes optional `nf1Variant` property to capture actual variant in accepted nomenclature (due to limitations/intention of current `nf1Genotype`). Currently `nf1Variant` has selections for common amino acid level changes (in an Asian population) curated from https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2733956/. When time permits, can go to the literature to build list with additional variants, e.g. also review (in an Italian population) https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4694136/ and use ClinVar, etc. 

Note: nf1Variant _could_ use a list rule.


allaway:
Hey, do you think the cell line/modelSystemName might be this one from Deanne Wallis? https://nf.synapse.org/Explore/Tools/DetailsPage/Details?resourceId=f047cff6-98c8-4562-b589-94b864ae35a1

NF.csv
```diff
@@ -750,9 +752,10 @@ parentSpecimenID,"A unique identifier (non-PII) that represents the parent speci
...
 nf2Genotype,"Genotype of NF2 gene in the biospecimen from which the data were derived, if known","-/-, +/-, +/+, +/?, -/?, Unknown",,FALSE,,neurofibromatosis,,,,nf2Genotype,Property,Biosample,annotationProperty,,Genotype,,
 nf1Genotype,"Genotype of NF1 gene in the biospecimen from which the data were derived, if known","-/-, +/-, +/+, +/?, -/?, Unknown",,FALSE,,neurofibromatosis,,,,nf1Genotype,Property,Biosample,annotationProperty,,Genotype,,
+nf1Variant,"Variant (mutation) in the NF1 gene/protein in the biospecimen from which the data were derived, if known/applicable.","R816X,E178X,L578P,R1276Q,R1513X,V1621R,R2237X,Y2264X6",,FALSE,,neurofibromatosis,,,,nf1Mutation,Property,Biosample,annotationProperty,,Genotype,,"See https://www.hgmd.cf.ac.uk/docs/mut_nom.html#general for nomenclature."
```
allaway, line 758:
```suggestion
nf1Variant,"Variant (mutation) in the NF1 gene/protein in the biospecimen from which the data were derived, if known/applicable.","R816X,E178X,L578P,R1276Q,R1513X,V1621R,R2237X,Y2264X6",,FALSE,,neurofibromatosis,,,,nf1Mutation,Property,Biosample,annotationProperty,,Genotype,,"See https://www.hgmd.cf.ac.uk/docs/mut_nom.html#general for nomenclature."
```

 I wonder if instead of tracking the variants here, we could reference the variant UUIDs from the NF Tools Database? What do you think? https://www.synapse.org/#!Synapse:syn26486835/tables/query/eyJzcWwiOiJTRUxFQ1QgKiBGUk9NIHN5bjI2NDg2ODM1IHdoZXJlIHByb3RlaW5WYXJpYXRpb24gPSAncC5Bcmc4MTZUZXInIiwgImFkZGl0aW9uYWxGaWx0ZXJzIjpbXSwgInNlbGVjdGVkRmFjZXRzIjpbXSwgImluY2x1ZGVFbnRpdHlFdGFnIjp0cnVlLCAib2Zmc2V0IjowLCAibGltaXQiOjI1LCAic29ydCI6W3siY29sdW1uIjoicHJvdGVpblZhcmlhdGlvbiIsICJkaXJlY3Rpb24iOiJERVNDIn1dfQ== (in this case, I believe we'd use the "transgenic" one)

allaway (changes requested):
In general LGTM, I just had a question about the cell line that was actually used in NFOSI-381 (I think we should still add vanilla HEK293T to the dictionary, so that's fine), and then whether we could use the NF Tools Database for variant tracking instead. 

Or, alternatively, maybe we should be using an external database like ClinVar to describe these variants, since we might need to describe patient variants for which there is no model organism (and thus we wouldn't have it in the tools DB).

anngvu:
> Hey, do you think the cell line/modelSystemName might be this one from Deanne Wallis? https://nf.synapse.org/Explore/Tools/DetailsPage/Details?resourceId=f047cff6-98c8-4562-b589-94b864ae35a1

Briefly skimming through their [protocol](https://www.synapse.org/#!Synapse:syn22322984), it's not clear whether it's created in-house or obtained from Wallis, so I'll check.

NF.csv
anngvu: Tbh, I think using UUIDs is going to be confusing and error-prone for contributors and not helpful if this is ever surfaced in the portal. We just don't have an interface where the label can be selected and the actual UUID is used for the index. Besides, there are only 3 variants tracked in the NF Tools Database currently. Moreover, the nomenclature is standardized enough that it will be easier to link than most things.

anngvu:
Confirmed this was not a cell line obtained from Wallis. However, they mentioned that they were using or will be using nine of Peggy Wallace's cell lines, so should try to have these linked for future annotation.

allaway:
Ah interesting! Thanks for letting me know! I will try to add this cell line to the tool database if they are willing to provide details.

anngvu:
@allaway Requested re-review for one question/discussion -- so I added a UUID column based on your review comment -- but this will also be in use for the cell line imports later on as well. In the DCA manifest, do you think we should allow selection from options with labels such as `R816X (transgenic)`, `R816X (spontaneous)`, etc., instead of just `R816X` then? This combines two concepts (the mutation + the mutation method) and I think we'll still need a simple `R816X` option when the method is unknown.

allaway:
>R816X (transgenic), R816X (spontaneous), etc., instead of just R816X then? This combines two concepts (the mutation + the mutation method) and I think we'll still need a simple R816X option when the method is unknown

TBH, I think that might be overly complicated, esp. if we are capturing the details elsewhere? I'd vote for just R816X.

allaway:
One more question actually, how do you envision the UUID column being used? Is this for _any_ database UUID?

anngvu:
> One more question actually, how do you envision the UUID column being used? Is this for _any_ database UUID?

@allaway 
1. I think you wanted to link variants and cell lines to the NF Tools Database UUIDs, so that is our source of UUIDs. 
2. Yeah, we can stick with just `R816X`. Not sure how closely you wanted to match the records in the Tools Database, which is why I proposed something like `R816X (transgenic)`. (I think this is more straightforward with the cell lines.)

allaway:
Thanks for clarifying! We probably then will need a mapping table that links UUIDs to concepts (since the NF Tools DB has lots of different UUIDs).

anngvu:
OK, so final changes: 

- Did a rebase just now since there some additions to Assay and Template modules today that resulted in `NF.csv` `NF.jsonld` conflicts
- Sticking with simple `R816X`/not going ahead with the transgenic options
- Not doing anything with UUID here, figured that's actually beyond scope of this PR and we can figure out changes later

allaway:
Sounds great, thanks!

allaway:
## Curation Review — NADIA Agent

**Synapse Project:** [syn74320111](https://www.synapse.org/#!Synapse:syn74320111)
**Source:** SRA:PRJNA897908
**PMID:** [36760809](https://pubmed.ncbi.nlm.nih.gov/36760809/)
**DOI:** 10.3389/fendo.2022.1070074

### Changes Applied

**Companion RNA-seq dataset** to syn74320102 (epigenomics). Same paper — PMID:36760809, Cascón lab CNIO.

**PMID and DOI added** — project was missing both.

**Annotation corrections:**
- `pmid` → added `36760809`
- `doi` → added `10.3389/fendo.2022.1070074`
- `studyLeads` → corrected to `Sara Mellid`, `Alberto Cascón`
- `institutions` → `Spanish National Cancer Research Centre CNIO`
- `fundingAgency` → `Medical Research Council`
- `dataType` → corrected to `geneExpression` (RNA-Seq; removed `genomicVariants`)
- `resourceStatus` → removed from 2 File entities

**File-level annotations:**
- `assay`: `rnaSeq`
- `species`: `Homo sapiens`
- `tumorType`: `Pheochromocytoma`
- `diagnosis`: `Neurofibromatosis 1`

**Files updated:** 2

Issue #230, opened 2023-01-03
Add back `isXenograft` column in RNAseq/Genomics Templates


jaybee84:
**Describe the bug**
The current manifest generated using the RNAseq or Genomics Template is missing `isXenograft` column

**To Reproduce**
1. Go to DCA
2. Click on Your favorite dataset
3. Scroll down to RNAseq template to generate manifest
4. See all columns but no `isXenograft` columns

**Expected behavior**
I would expect to see a column named `isXenograft` similar to `isCellLine` or `isPrimaryCell`

**Priority** (select one)
- [ ] Minor ⬇️
- [ ] Major 📢
- [x] Critical 🆘

**Additional context**
This information is critical to have for future data processing and disambiguation.

PR #235, opened 2023-01-11
add add'l xenograft terms

allaway:
(transplantationType,transplantationRecipientSpecies,transplantationRecipientTissue) to templates. Caused by #230.


anngvu (approved):
Thanks for fix. Though I think maybe what I should have done is have`isXenograft` set up with these addt'l as dependencies so they're always brought along if this attribute is included (similarly to `dataType` bringing along `dataSubtype`).

PR #237, opened 2023-01-20
Patch/cell lines toolsdb

anngvu:
This pulls in new cell lines seen in tools db. The sync script gives general idea and can be updated for mouse models, etc. (as separate PR). 

One question: Using the `resourceName` right now -- though eventually use `resourceIds`?
In Tools DB, [iPSC NF1 +/- BJFF.6 bkgd has two different `resourceIds`](https://www.synapse.org/#!Synapse:syn26438037/tables/query/eyJzcWwiOiJTRUxFQ1QgKiBGUk9NIHN5bjI2NDM4MDM3IHdoZXJlIHJlc291cmNlSWQgaW4gKCdhMGM5MTYyNy1mYWQ2LTRkNjAtYmY3NC02NzYyMzM3NGVmZjgnLCAnZTk4MTc5NmUtMTRkOC00ZDAxLWE1MDAtNGFlMjNmMDY2MGJlJykiLCAiYWRkaXRpb25hbEZpbHRlcnMiOltdLCAic2VsZWN0ZWRGYWNldHMiOltdLCAiaW5jbHVkZUVudGl0eUV0YWciOnRydWUsICJvZmZzZXQiOjAsICJsaW1pdCI6MjV9) (maybe based on the different `donorId`, though `parentDonorId` is the same). Wanted to make sure this was correct.


allaway:
I think it's fine to use resourceName for now. One of the problems with that is that there are some very generic names (eg cell line named NF1) so there could be collisions in the future. But as we previously discussed elsewhere, the UUID approach is not very easy to use from the annotation side.

> iPSC NF1 +/- BJFF.6 bkgd has two different ids
Thanks for mentioning that .... I'll take a look. I think that's probably a bug.

allaway (approved):
LGTM. One thought, instead of Mouse Model maybe could we classify them as "Animal Models" to have consistency with the NF tools db?

Issue #254, opened 2023-03-13
Add registered JSON schemas 


anngvu:
These JSON schemas are part of the NF (meta)data model. They are registered with the Synapse platform and can be bound to folders or files. Currently, they cannot be generated from schematic/are outside of schematics scope and must be composed separately. 

Currently, we have one schema bound to the "Raw Data" folder to indicate that child folders contain "datasets" (i.e. the set of files in each child folder is conceptually a dataset).

PR #255, opened 2023-03-13
Feat/registered schemas

anngvu:
See #254. This is the schema that **will be bound to already-created "Raw Data" folders** with `enableDerivedAnnotations = true`. 
To see what it looks like, go to:

- [Raw Data (test)](https://www.synapse.org/#!Synapse:syn51106349)
- Create a new folder under Raw Data, e.g. with name "Bonus dataset"
- Go to "Bonus dataset" entity page. Click on the "Annotations" icon. It will say there are no annotations. However, click **Edit**. Note that contentType is automatically set. You can also select an assay value from the drop-down. 
- Creating a nested folder under "Bonus dataset" will work the same.

This is highly relevant to issue https://github.com/nf-osi/jobs/issues/50. I think for _new_ projects, contentType and assay will be set directly as annotations instead of binding the "superdataset" schema to Raw Data to get recursive derived annotations. (Leave comments at https://github.com/nf-osi/jobs/issues/50 if there are other thoughts...)


allaway:
Hmm, I just tried to test this and I did not see anything pop up when I clicked "edit" on the annotations panel. Am I doing something wrong?

anngvu:
@allaway Sorry, forgot that I had Experimental mode = On, which is default for me. The experimental mode allows you to see the derived annotations that don't show up in normal UI (but they'll show up in a fileview no matter what mode).

allaway (approved):
LGTM - reviewed in 1-1!

PR #281, opened 2023-04-13
Patch for NFOSI-521

allaway:
In working on NFOSI-521, I discovered that photograph was not being surfaced in the manifest even though we had it as an assay term. This corrects that.


anngvu (approved):
:+1:

PR #286, opened 2023-05-11
Update NF.csv

cconrad8:


cconrad8:
fixing single , that prevented table from formatting correctly.

PR #287, opened 2023-05-12
changes to metadata

cconrad8:


anngvu:
@cconrad8 I wrote a linter with your changes to see where there may be a formatting issues. Usually the most common problems (which I've experienced!) are:
- Wrong number of fields of commas, not quoting etc
- Missing return at the end of file.

I will branch off of your branch on this and add the linter to see if it can automatically fix all the issues so you don't have to redo anything.

cconrad8:
> @cconrad8 I wrote a linter with your changes to see where there may be a formatting issues. Usually the most common problems (which I've experienced!) are:
> 
> * Wrong number of fields bc commas, not quoting etc
> * Missing return at the end of file.
> 
> I will branch off of your branch on this and add the linter to see if it can automatically fix all the issues so you don't have to redo anything.

Ok thank you! @anngvu I was looking at this earlier and used Notepad ++ which has a csv linter plug in....but I couldn't find anything wrong still...!

PR #292, opened 2023-05-18
Patch/new terms with linter

anngvu:
Part 1 of 3 in workflow updates to make life easier/reduce those red Xs; this one adds the linter.

https://www.freecodecamp.org/news/what-is-linting-and-how-can-it-save-you-time/
> ...a linter is a tool that programmatically scans your code with the goal of finding issues that can lead to bugs or inconsistencies with code health and style. Some can even help fix them for you! 

I just added the linter on top of the changes in https://github.com/nf-osi/nf-metadata-dictionary/pull/287. Go to **Files changed** tab when the check seems to fail. It will try to show where:

- A file has been auto-corrected for line endings ([someone's explanation](https://gist.github.com/camh-/1bebfcff1b0f814e9b191edc60d5206b) of line endings) (before it is corrected, GitHub will mark with red circle in the file comparison)
image

- A line that doesn't look as expected (mismatched field numbers)
image


modules/Data/File_Format.csv
cconrad8, line 1: @anngvu I can't tell - what was changed here?

Issue #297, opened 2023-05-23
Add new file formats from incoming data


anngvu:
**List the new value**
Value: 
- `docker image`
- `ome-tiff`

**Additional comments**
It looks like these were added in dcc-site schema initially without being added here.

allaway:
Tagging onto this issue - change hdf file format to hdf5 to be more explicit. There are a small subset of current files that are annotated hdf (with file extension h5) that need to be changed with this, I'll change those annotations to hdf5 now.

PR #304, opened 2023-05-26
address #297

allaway:
#297


allaway:
Will trigger a new MAJOR version as hdf switched to hdf5 (MAJOR: In-use parent attributes are deleted from dictionary or modified, or in-use child attributes are modified in a non-backwards compatible way (e.g. Neurofibromatosis 1 changed to Neurofibromatosis type 1).)

anngvu (commented):
New values not added to Valid Values under fileFormat, but this will be fixed in next update

PR #305, opened 2023-06-12
Improve testing development

anngvu:
Testing and development improvements factored out from the LinkML/auto-values feature branch to be reviewed first because this makes more sense and will be easier to review. (Without splitting into multiple smaller PRs, there are 78 changed files...)

These changes prep for that next PR by creating an environment ready for LinkML/easier development overall, and by ensuring the correctness of the new implementation with a better test suite.


anngvu:
@allaway I will be removing "GenomicsAssayTemplateExtended" anway because it's not being used and too complicated, so you can ignore that result.

For validation, test manifest 4 is supposed to give error but only gives warning. Test manifest 5 is schematic definitely failing because validations thinks 0 is OK...

Issue #277, opened 2023-04-11
Make annotations required for RNA sequencing 


cconrad8:
Change the following annotations to be required for RNA sequencing: 

https://github.com/nf-osi/nf-metadata-dictionary/blob/e7268f2ef696c60b75a121f20790473d2ec464c8/docs/templates/RNASeq_Template.csv

individualID
specimenID 
assay 
species
libraryPreparationMethod
Platform
readPair
specimenPreparationMethod
tumorType

This would help us eliminate the need to ask people afterwards for the annotations. It would allow users to re-process the data including us for NTAP projects.  

Potentially also do for WES and WGS

allaway:
I agree with this. However, I have a thought about readPair - this one might not apply in all cases. For example, if the data are in BAM or CRAM format, this is not applicable. I don't think we should add "not applicable" as a value, because it adds unnecessary annotation burden rather than just leaving it blank. Also, this one is almost always possible to infer from the name (e.g. the filename usually ends in _R1 or _1,  _R2 or _2), so we can add those annotations without reaching out to the PI.

cconrad8:
@allaway agree with you on "read pair" not being included.  Thanks!

anngvu:
Notes:
- `assay` and `platform` already required, e.g. https://github.com/nf-osi/nf-metadata-dictionary/blob/main/NF.jsonld#L4810
- not requiring `readPair`

Issue #118, opened 2022-03-03
Add typing using Valid Values field


anngvu:
Except for 1-2 keys, Validation Rules column is not used. Aside from better validation, availing this field for formally specifying types can help relate the schema defined in the schematic framework with the fileview schema in Synapse, e.g. through mapping of "list" to "StringList", and "int" to "Integer". See https://github.com/nf-osi/jobs/issues/10. 

Note that some things allowed for Validation Rules (regex expressions) doesn't really make sense as a Synapse type, so any future regex will have to be handled/ignored when mapping.

PR #318, opened 2023-07-13
Patch/strict model

anngvu:
WIP - todo - add new tests.

The theme of this PR is making a stricter model, relevant to #277 and #118. There might be additional fields that could use more enforcement, i.e. this might not be exhaustive but should still crack down on many more than previously.


anngvu:
Update: found an issue that maybe explains why some required columns are not highlighted blue: https://github.com/Sage-Bionetworks/schematic/issues/1018

cconrad8 (approved):
Looks good to me! Thanks for making the changes

Issue #339, opened 2023-09-05
Add constraints to match template to assay/datatype


anngvu:
Currently, it's pretty easy for a data contributor to select the wrong template for the assay/datatype, which is understandable because of the long selection list and lack of guidance at the interaction point (DCA UI).

We can use these rules/constraints not just for validation but also for materializing guidance, tagging dataset folders with the correct template. In the future, this can also be picked up in the app to select the correct template automatically.

Issue #338, opened 2023-08-31
Refactor templates to use class inheritance


anngvu:
This is currently being tested in the Gray Foundation model first, but it is a nice modeling feature akin to having dry enums. It would allow neater, more understandable templates because properties (slots) would be automatically expanded according to the class lineage, so e.g. scRNASeqTemplate would inherit sequencing relevant props from SequencingTemplate, and the scRNAseqTemplate should only contain props that are specific to single-cell sequencing.

Issue #251, opened 2023-03-07
data type required when not applicable


allaway:
**Describe the issue with the metadata**
Data type might not be applicable for some resource types, e.g. 'report', but it's still required (and there's no "not applicable" value) 

**List alternative**
n/a - should remove as requirement IMO

**Additional comments**

Issue #168, opened 2021-04-15
Break genomics assay templates into levels like HTAN


kthrog:
**Is your feature request related to a problem? Please describe.**
Suggested to reduce annotation burden. 

**Describe the solution you'd like**
Break genomics assay template in NF Data Curator app into two levels: 
- Level 1 for unaligned and aligned reads
- Level 2 for counts 

**Describe alternatives you've considered**
n/a

**Additional context**
Based on HTAN's levels, documented here: https://docs.google.com/document/d/1dJ7NUoVCtewdtny8bITwtWnzItB4IibL5kJO3ZNh0go/edit#heading=h.aia8yotky7pe

PR #342, opened 2023-09-08
Feat/advanced templates

anngvu:
- #338. Why:
  - This is **an under-the-hood change that allows cleaner and more consistent reusable building blocks**. I think I tried to do this within what schematic provided, but it didn't work as well as I'd like and felt like a mistake. This is conceptually nicer. 
  - **Easier to compare templates and see how they are different/specialized**. See more meaningful relationship between templates through the inheritance hiearchy.
  - One **possible cons** is that because it is easier to make templates now, we'll have **over-proliferation** of templates (some more specific than we need, so need to keep that in mind.)
- #251 / #167. Why:
  - Yes, just don't have these in non-applicable templates (like for protocols). Looks more obvious what's happening now.
- #339. Why:
  - While **technically not able to be used by schematic yet _in the future_** it might actually be useful for improved UI and validation. 
  - **Outside of schematic, *we* can possibly use these constraints to make recommendations on Synapse/in custom validation workflow (nfportalutils)**. If we only know about assay, the default is to suggest the raw data template for that assay. If we know both assay and datatype (ultimately, datatype is more important), we can do a better job of directing contributors to the most appropriate template 🥺 
  - **Additional clarity**, helps to realize whether or not all expected assays/datatypes **are actually accounted for** in the available template selection...
- #114 / #168. Why:
  - Technically, we already started doing this for NF-processed data. Having these mappings/alignments is helpful and can improve quality. 
  - Perhaps provide starting example for cross-team data model mapping.
- #138 -> "libraryStrand". Why:
  - This was just overdue, but this change will "improve quality".
- #343. Why:
  - Assurance that everything still works with latest stable version of schematic!

**Update:** 
- Added tests, not sure if Christina has gotten a pre-official-review time as a heads-up for how proposed changes might fit into this, but I'll schedule some time for interactive working session as well where we can experiment with a fork of this.


modules/Assay/Assay.yaml
```diff
@@ -9,6 +9,8 @@ enums:
         meaning: https://www.thermofisher.com/us/en/home/life-science/cell-analysis/fluorescence-microplate-assays/microplate-assays-cell-viability/alamarblue-assay-cell-viability.html
       2D CellTiter-Glo:
         description: Cell viability assay based on detection of ATP.
+        todo:
+        - This should be subsumed under `cell viability assay`, can add CellTiter-Go as platform.
```
allaway, line 13: Maybe same for 2D incucyte below?

modules/Assay/Parameter.yaml
```diff
@@ -44,111 +17,153 @@ enums:
...
-        description: Fluorescent dye used for the red channel on an array
-        meaning: http://purl.obolibrary.org/obo/FBbi_00000450
-      Cy3:
-        description: Fluorescent dye used for the green channel on an array
-        meaning: http://purl.obolibrary.org/obo/FBbi_00000449
-  GenePerturbationTechnologyEnum:
+      singleEnd:
+        description: A library preparation that results in the creation of a library of 5' ends of DNA.
+        meaning: Sage Bionetworks
+  StrandednessEnum:
```
allaway, line 155: Do you think it is worth making this identical to HTAN's implementation of this concept? 
eg

term: Library Strand
values: Unstranded, FirstStranded, SecondStranded, NotApplicable

modules/Assay/Assay.yaml
anngvu: Yeah, agree

modules/Assay/Parameter.yaml
anngvu: Done

PR #346, opened 2023-09-22
Add new PRO assays and data types

allaway:
Needed to merge this DSP: https://github.com/nf-osi/dcc-site/pull/659

### DIFF
--- a/modules/Assay/Assay.yaml
+++ b/modules/Assay/Assay.yaml
@@ -360,7 +360,7 @@ enums:
         see_also:
           - Dynamic light scattering (DLS) is the ancestor of microrheology.
         source: https://www.elveflow.com/microfluidic-reviews/general-microfluidics/microrheology-a-review/
-      modified Skindex:
+      Skindex-16:
         description: A validated measure of the effects of skin diseases on quality of life that [is] suitable for use in research about patients' experiences of illness and its treatment
         source: https://pubmed.ncbi.nlm.nih.gov/22284137/
       multi-electrode array:
--- a/modules/Assay/Assay.yaml
+++ b/modules/Assay/Assay.yaml
@@ -83,6 +83,11 @@ enums:
           An imaging-based system that allows measurement and analysis of peaks of human-derived induced pluripotent stem cells (hiPSCs), 
           differentiated into cardiomyocytes and neurons,  up to 100 times per second and quickly cherry pick events such as Early-After-Depolarization-like events (EAD-like events), fibrillation, and irregular beating.
         source: https://www.moleculardevices.com/products/flipr-penta-high-throughput-cellular-screening-system
+      Focus group:
+        description: > 
+          A small, usually diverse group of people whose response to something is studied to determine the response that can be expected from a larger population. 
+          It is used especially in market research and political analysis.
+        source: http://purl.obolibrary.org/obo/NCIT_C154589
       FTIR spectroscopy:
         description: > 
           Fourier transform infrared (FTIR) spectroscopy is a technique used to obtain an infrared spectrum of absorption or emission of a solid, liquid or gas. 
@@ -219,6 +224,9 @@ enums:
       conventional MRI:
         description: Magnetic resonance imaging using standard protocols for high resolution structural and anatomic characterization, including T1 weighted, T2 weighted, fluid attenuated inversion recovery (FLAIR), and gadolinium-enhanced sequences. [  NCI  ]
         meaning: http://purl.obolibrary.org/obo/NCIT_C175525
+      Children's Dermatology Life Quality Index Questionnaire:
+        description: A standardized rating scale originally developed by Lewis-Jones and Finlay in 1995. This instrument is used to make quality of life assessment in children with skin conditions. [ NCI ] 
+        meaning: http://purl.obolibrary.org/obo/NCIT_C119092
       differential scanning calorimetry:
         description: Differential scanning calorimetry (DSC) is the measurement of thermodynamic parameters (e.g. enthalpy) during a chemical or biochemical reaction of both a sample and a reference, by the known variation (step-wise or linear) of one variable, whilst a second is kept constant.
         meaning: http://purl.obolibrary.org/obo/CHMO_0000684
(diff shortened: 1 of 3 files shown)


modules/Data/Data.yaml
```diff
@@ -91,6 +91,14 @@ enums:
       electrophysiology:
         description: Data generated from an electrophysiology assay.
         meaning: http://purl.obolibrary.org/obo/ERO_0000564
+      Patient Reported Outcome Measures:
```
anngvu, line 94: Hmmmmm, I recommend using the existing `surveyData`. It can be tricky when to know how to break out a concept, but if I were doing reporting this would get merged with `surveyData` anyway.

anngvu (changes requested):
Reluctant on some data type additions, maybe these can be revised.

modules/Data/Data.yaml
allaway: Yeah, that's a fair point!

modules/Data/Data.yaml
```diff
@@ -91,6 +91,14 @@ enums:
       electrophysiology:
         description: Data generated from an electrophysiology assay.
         meaning: http://purl.obolibrary.org/obo/ERO_0000564
+      Patient Reported Outcome Measures: 
+        description: Survey measures that are completed by the patient which help assess patient status with regards to pain, mobility, quality of life, ability to perform daily tasks or notable events in a clinical study. [ NCI ] 
+        meaning: http://purl.obolibrary.org/obo/NCIT_C177377
+      demographics: 
+        description: The statistical characterization of human populations or segments of human populations (e.g., characterization by age, sex, race, or income).
+        meaning: http://purl.obolibrary.org/obo/NCIT_C16495
+      audio transcript:
```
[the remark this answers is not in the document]
allaway: Yeah, this was a tough one for me too. I was going to just put "transcript" but then realized that there's the possibility of confusion with the biological type of transcript...

modules/Data/Data.yaml
allaway:
```suggestion
```

modules/Data/Data.yaml
```diff
@@ -91,6 +91,14 @@ enums:
       electrophysiology:
         description: Data generated from an electrophysiology assay.
         meaning: http://purl.obolibrary.org/obo/ERO_0000564
+      Patient Reported Outcome Measures: 
+        description: Survey measures that are completed by the patient which help assess patient status with regards to pain, mobility, quality of life, ability to perform daily tasks or notable events in a clinical study. [ NCI ]
```
allaway, line 95:
```suggestion
```

modules/Data/Data.yaml
```diff
@@ -91,6 +91,14 @@ enums:
       electrophysiology:
         description: Data generated from an electrophysiology assay.
         meaning: http://purl.obolibrary.org/obo/ERO_0000564
+      Patient Reported Outcome Measures: 
+        description: Survey measures that are completed by the patient which help assess patient status with regards to pain, mobility, quality of life, ability to perform daily tasks or notable events in a clinical study. [ NCI ] 
+        meaning: http://purl.obolibrary.org/obo/NCIT_C177377
```
allaway, line 96:
```suggestion
```

modules/Data/Data.yaml
allaway, line 94:
```suggestion
```

modules/Assay/Assay.yaml
```diff
@@ -338,6 +357,8 @@ enums:
         see_also:
           - Dynamic light scattering (DLS) is the ancestor of microrheology.
         source: https://www.elveflow.com/microfluidic-reviews/general-microfluidics/microrheology-a-review/
+      modified Skindex:
```
jaybee84, line 360: Should we add Skindex too? In looking through I only found skindex mini : https://bioportal.bioontology.org/ontologies/LOINC?p=classes&conceptid=http%3A%2F%2Fpurl.bioontology.org%2Fontology%2FLNC%2FLP420590-4

modules/Data/Data.yaml
```diff
@@ -91,6 +91,11 @@ enums:
       electrophysiology:
         description: Data generated from an electrophysiology assay.
         meaning: http://purl.obolibrary.org/obo/ERO_0000564
```
jaybee84, line 93: I am getting a "Term not found" error using the link :http://purl.obolibrary.org/obo/ERO_0000564 

Since there are multiple electrophysiology assays, it may be time to update this to be more granular.

allaway: Since this term is out of scope for this PR, could you add this as an issue? BTW, this is for the dataType module, not the assay module - so maybe it's fine to have the broader term here?

modules/Assay/Assay.yaml
allaway: Skindex was found to be insufficient for use in NF patients. I am guessing we do not need to add it because people will probably not use the normal skindex. 

There are several "modified" flavors of it (e.g. cNF-skindex). I'm not sure if any of the modified flavors are available in any ontologies.

allaway: I will check with Sang to see if these investigators are using cNF skindex or something else.

allaway, line 360:
```suggestion
     Skindex:
```

modules/Assay/Assay.yaml
```diff
@@ -338,6 +357,8 @@ enums:
...
+        description: ''
```
allaway, line 361:
```suggestion
        description: '[a validated measure] of the effects of skin diseases on quality of life that [is] suitable for use in research about patients’ experiences of illness and its treatment'
        meaning: https://pubmed.ncbi.nlm.nih.gov/22284137/
```

modules/Assay/Assay.yaml
allaway: Turns out they are using cNF-skindex and skindex-16 for this study, so I got rid of the "modified" and added these two.

modules/Data/Data.yaml
```diff
@@ -91,6 +91,11 @@ enums:
...
+      demographics:
```
allaway, line 94:
```suggestion
```

modules/Data/Data.yaml
```diff
@@ -91,6 +91,11 @@ enums:
       electrophysiology:
         description: Data generated from an electrophysiology assay.
         meaning: http://purl.obolibrary.org/obo/ERO_0000564
+      demographics: 
+        description: The statistical characterization of human populations or segments of human populations (e.g., characterization by age, sex, race, or income).
```
allaway, line 95:
```suggestion
```

modules/Data/Data.yaml
```diff
@@ -91,6 +91,11 @@ enums:
...
+        meaning: http://purl.obolibrary.org/obo/NCIT_C16495
```
allaway, line 96:
```suggestion
```

anngvu (approved):

Issue #351, opened 2023-10-11
Add cram, crai, csi, tbi formats to fileFormat


allaway:
**List the new value**
Value: `cram`

**Provide definition and source for the new value**
Definition: A CRAM file is a compressed format for storing genomic sequence data. 
Source: ChatGPT (September 25, 2023 Version)

**List the new value**
Value: `crai`

**Provide definition and source for the new value**
Definition: A CRAI file is an index for a CRAM file, facilitating fast data retrieval.
Source: ChatGPT (September 25, 2023 Version)

**List the new value**
Value: `csi`

**Provide definition and source for the new value**
Definition: A CSI file is a compressed sequence index used for efficiently accessing genomic data in large datasets.
Source: ChatGPT (September 25, 2023 Version)

**List the new value**
Value: `tbi`

**Provide definition and source for the new value**
Definition: A TBI file is an index for a TABIX-compressed genomic data file, enabling rapid region-based retrieval.
Source url: Source: ChatGPT (September 25, 2023 Version)

PR #353, opened 2023-10-12
Issue #351

allaway:
Ok, so my source might be a bit controversial, but I figure it's just as good (if not better) than the "Sage Bionetworks" items. I couldn't find any of these terms in OLS4 and the spec sheets left me wanting something more pithy and informative.

### DIFF
--- a/modules/Data/FileFormat.yaml
+++ b/modules/Data/FileFormat.yaml
@@ -102,6 +102,15 @@ enums:
       cns:
         description: Segmented log2 ratios from CNVKit pipeline, these represent the most relevant output describing copy number variation. 
         source: https://cnvkit.readthedocs.io/en/stable/fileformats.html
+      cram:
+        description: A CRAM file is a compressed format for storing genomic sequence data.
+        meaning: ChatGPT (September 25, 2023 Version)
+      crai:
+        description: A CRAI file is an index for a CRAM file, facilitating fast data retrieval.
+        meaning: ChatGPT (September 25, 2023 Version)
+      csi:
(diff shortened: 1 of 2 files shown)


modules/Data/FileFormat.yaml
```diff
@@ -102,6 +102,15 @@ enums:
       cns:
         description: Segmented log2 ratios from CNVKit pipeline, these represent the most relevant output describing copy number variation. 
         source: https://cnvkit.readthedocs.io/en/stable/fileformats.html
```
allaway, line 104: @anngvu what is the difference between "meaning" and "source"? Should I be using one or the other?

anngvu: `meaning` should be an ontology uri. `source` is for some reference (so "Sage Bionetworks", maybe even "James Eddy", but most commonly some website or paper)

anngvu: I'll add this topic to the much-needed doc update...

allaway: Ok! I'll change this to "source" where relevant.

anngvu: So I don't think I completely segregated our original `source` values for terms into LinkML's particular of `meaning` vs `source` when I did the migration to LinkML, so this isn't too important for now. I think we can go back and clean this up when we do need to make sure that `meaning` are valid uris.

anngvu:
Thanks, I think we can go back to move the `meaning` later to `source` since there are some terms where this needs to be refined as well.

anngvu (approved):

PR #358, opened 2023-10-17
Patch templates props

cconrad8:


cconrad8:
I'm not sure if specimenID should be required now..I've since come across situations where it's not surfaced and I'm not sure what to put in that situation.

cconrad8:
@anngvu Yeah I think i was the one that requested it to be made required before.... 😬
It's for this project: https://docs.google.com/spreadsheets/d/1rFuEeLYO-WNeVOBppb3TYaLm7YO1Pe1PMPC57hAPlIw/edit
The excel files don't have IDs

Is there a way to make the terms required in one but not required in the other?

anngvu:
Hmm, one of the files is not really raw data -- just summarized data with bar charts -- I think we should move syn52522934 to  Analysis since it's not really meant for the Imaging Assay template. The other [syn52522936](https://www.synapse.org/#!Synapse:syn52522936) looks more like clinical results data (REDCAP export). So I think we can use the Clinical Assay here, which doesn't ask for specimenID. But still need individualIDs. Is it possible to ask them about the deidentified ids associated with this data?

modules/props.yaml
```diff
@@ -629,7 +629,7 @@ slots:
     description: > 
       A unique identifier (non-PII) that represents the subspecimen (subsample) from which the data came, 
       e.g. an ID that distinguishes between different parts of the same parent tumor specimen.
-    required: true
+    required: false
```
anngvu, line 632: Change back to `true`.

anngvu (changes requested):

modules/props.yaml
cconrad8: this has been changed back to true

allaway:
LGTM!

allaway (approved):

anngvu (approved):

PR #360, opened 2023-10-19
Patch/docs for linkml

### LINKED ISSUE #340: Update docs + add LinkML specifications/mapping
Much of our README needs to be revised... Also translate specifications of how schematic maps to LinkML as somewhat discussed with Milen, Pierrette, and Tim Putman.

anngvu:
- Close #340 

@cconrad8 recently updated the README for me, ha, so this is mostly focused on the LinkML and other details here and there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,114 @@
 
 ### Maintenance and Contribution
 
-The purpose of the metadata dictionary is to provide a set of standard terms to describe data. Terms in the metadata dictionoary are used in the manifests within the [data curator app](https://dca.app.sagebionetworks.org/). 
-
+The purpose of the data model is to provide a set of standard concepts and structure that help to describe and manage data. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,23 +2,142 @@
...
+    range: TissueEnum # Take a look at TissueEnum below
+    required: false
+  title:
+    description: Title of a resource.
+    meaning: https://www.w3.org/TR/vocab-dcat-3/#Property:resource_title
+    required: true
+  # ...more below
+```
+
+Here, `required` defines globally whether this slot *must* be filled out.
```
cconrad8, line 83: perhaps here we can add if you need the case where it's not required globally what to do (if there is a solution)

cconrad8 (approved):

README.md
anngvu: That's a good suggestion

PR #372, opened 2023-11-30
Change Tissue to SpecimenType

allaway:
This is a breaking (major) change. Changes the "tissue" term to "specimenType". To avoid re-mapping, I've left the current "tissue" terms as-is, and just added new ones that we need like 'saliva. '


anngvu (approved):
I think everything should be good for this update (including the docs).

Issue #371, opened 2023-11-30
assayTarget not surfaced in imaging assay template


allaway:
**Describe the issue with the metadata**
This attribute is to collect the antibody targets of a IHC, IF, FACS, or other similar assays. It's possible that assayTarget is too vague. Maybe we want something more descriptive, so that it is easier for people to understand what we are collecting here. We could also create an IHC/IF-specific template so that we don't surface this term in the generic Imaging template.

Issue #370, opened 2023-11-29
Add the` dca-template-config.json` to this repo and use for testing


anngvu:
Will likely bundle this with the next data model update, similarly to GF's PR [here](https://github.com/gf-dcc/data-model/pull/80):

In [this Slack announcement](https://sagebionetworks.slack.com/archives/C01ANC02U59/p1700177506633809), D&T announced that we can maintain the templates config.json somewhere and DCA will reference the link (probably because keeping on top of PRs to update configs for 10+ DCCs is a lot of work):

>     For the next release of DCA in mid-December we are going to switch to new config files
>     ...
>     the template dropdown config and logo will be referenced by a URL, not the filepath in data_curator_config. This means those files can live elsewhere, like your data model repo.

Issue #369, opened 2023-11-28
Add quantseq prep kits to libraryPrep


allaway:
**List the new value**

Just taking notes during a meeting so don't have time to fill out the template, but Katherine Piculell is using this prep kit : https://www.lexogen.com/quantseq-fwd-udi-v2/ for some of their low-abundance samples. This is pretty important to track as it only provide gene-level counts, not transcript-level. Still could be useful data, but need to have this in mind when analyzing.

Value: `type value here`

Using [OLS](https://www.ebi.ac.uk/ols/index) or a similar resource, please find a standard term that fits your needs. OLS-listed NCIT and OBI dictionaries are preferred sources. Non-OLS sources of ground truth for specific items such as ONCOTREE for cancer types, or GEO for platform definitions are also appropriate. 

**Provide definition and source for the new value**
Definition:
Source url:

**Are there similar but not-quite-right values?**
Please list if applicable.

**Additional comments**

PR #375, opened 2023-12-04
Patch/fixes and reconfig

anngvu:
Issues:
- #371 - I think we lost `assayTarget` in the attempt to do more granular refactoring. Please see my own thoughts below and let me know what you think. 
- #370 - See issue + we need to point staging to config added here per D&T [notification](https://sagebionetworks.slack.com/archives/C01ANC02U59/p1701367488599259)
- #369 - Getting out of the way since an easy addition

### DIFF
--- a/modules/Assay/Parameter.yaml
+++ b/modules/Assay/Parameter.yaml
@@ -3,33 +3,33 @@ enums:
     permissible_values:
       Cy3:
         description: Fluorescent dye used for the green channel on an array
-        source: http://purl.obolibrary.org/obo/FBbi_00000449
+        meaning: http://purl.obolibrary.org/obo/FBbi_00000449
       Cy5:
         description: Fluorescent dye used for the red channel on an array
-        source: http://purl.obolibrary.org/obo/FBbi_00000450
+        meaning: http://purl.obolibrary.org/obo/FBbi_00000450
   GenePerturbationEnum:
     permissible_values:
       knockdown:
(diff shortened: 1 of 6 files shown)


modules/Assay/Parameter.yaml
```diff
@@ -3,33 +3,33 @@ enums:
...
       CRE Recombinase:
         description: CRE Recombinase catalyses site-specific recombination between two 34 base pair loxp sites and maintains the phage genome as a monomeric unit-copy plasmid in the lysogenic state.
-        meaning: http://purl.obolibrary.org/obo/NCIT_C17285
+        source: http://purl.obolibrary.org/obo/NCIT_C17285
       CRISPR:
         description: ''
       RNAi:
         description: High throughput sample analysis of RNAi molecules for potential application in gene knockdown or gene silencing of target genes
-        meaning: http://purl.obolibrary.org/obo/ERO_0001688
+        source: http://purl.obolibrary.org/obo/ERO_0001688
```
allaway, line 32:
```suggestion
        source: http://purl.obolibrary.org/obo/ERO_0001688
```

I'm confused - I thought that for pURLs, we should use `meaning` and for other sources use `source` - did I misunderstand? Or were these changed to source by accident?

allaway: https://github.com/nf-osi/nf-metadata-dictionary/pull/353/files#r1357085894

allaway (changes requested):
I had one comment - not sure if it's really requesting changes, more of a clarification about something I possibly misunderstood.

allaway (approved):
LGTM!!

PR #380, opened 2023-12-08
add Safer-SeqS

allaway:
This is currently a placeholder definition, because the SOW we reviewed said "we developed this assay..." but when I look for publications there are several different groups (none of which involve the relevant PI) that have developed a method called something like "Safer-SeqS". we should get the correct citation and add it here!

### DIFF
--- a/modules/Assay/Assay.yaml
+++ b/modules/Assay/Assay.yaml
@@ -458,8 +458,9 @@ enums:
       scale:
         description: An instrument or machine for weighing.
         meaning: http://purl.obolibrary.org/obo/MMO_0000217
-      Safer-SeqS:
+      SaferSeqS:
         description: A blood-based assay that evaluates mutations in circulating tumor DNA and chromosomal copy number changes to detect minimal residual disease (MRD) with high specificity towards tumor type.
+        source: https://doi.org/10.1038/s41587-021-00900-z
       single molecule drug screen assay:
         description: An experiment in which a single molecule was used in an assay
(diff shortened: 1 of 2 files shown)


jaybee84:
Yikes! I was working on a PR for this too :D

modules/Assay/Assay.yaml
```diff
@@ -458,6 +458,8 @@ enums:
       scale:
         description: An instrument or machine for weighing.
         meaning: http://purl.obolibrary.org/obo/MMO_0000217
+      Safer-Seq:
+        description: A blood-based assay that evaluates mutations in circulating tumor DNA as well as chromosomal copy number
```
jaybee84, line 462:
```suggestion
        description: A blood-based assay that evaluates mutations in circulating tumor DNA and chromosomal copy number changes to detect minimal residual disease (MRD) with high specificity towards tumor type.
```

modules/Assay/Assay.yaml
```diff
@@ -458,6 +458,8 @@ enums:
       scale:
         description: An instrument or machine for weighing.
         meaning: http://purl.obolibrary.org/obo/MMO_0000217
+      Safer-Seq:
```
jaybee84, line 461:
```suggestion
      Safer-SeqS:
```

jaybee84 (approved):

allaway:
Ah, haha, sorry! See my comment on Slack - I am going to change this to SaferSeqS

PR #385, opened 2023-12-18
Split out main processed data vs merged processed data

anngvu:
For merged data files that aggregate samples into one file, we're just going to be on par with GDC, see:
```
close_mappings:
    - https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=aggregated_somatic_mutation
    - https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=gene_expression
```

### DIFF
--- a/NF.jsonld
+++ b/NF.jsonld
@@ -6445,7 +6445,7 @@
     "sms:required" : "sms:false",
     "sms:requiresComponent" : "",
     "rdfs:label" : "ProcessedExpressionTemplate",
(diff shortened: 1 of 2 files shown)


cconrad8 (approved):
looks good to me 

writing down just so we have on record - that a description of the templates would be helpful for the DCA to surface since these manifest titles are very similar

Issue #334, opened 2023-08-21
Conceptualize normalized clinical data/sample-centric template


anngvu:
**Update** 
- Ready for officially collecting comments and ideas on Sep 26, 2023, tagging @allaway @jaybee84 @cconrad8.
- Scheduling for Sprint 25 so ~3 weeks for gathering comments.

***

Notes and discussion for what a clinical-focused sample-centric table would look like. 

Initial thoughts:

### Use cases 
- Cleaner way to get information on # of individuals and samples for data in the portal (this can be somewhat be achieved with file metadata, but not as cleanly/requires data wrangling).  
- Help with data-entry errors in several steps of the process: Use as a reference for validation. Or use as a reference to simplify annotations needed during file annotation (which can be bad for users if there are many files). 
- Through this, allow users to contribute more useful attributes other than basic ones of `sex`, `age`, etc. "Minimal" clinical data defined in [Table 2](https://dash.harvard.edu/bitstream/handle/1/32725809/Neurology-2016-Hanemann-S40-8.pdf?sequence=1&isAllowed=y) here, which some groups also wanted to extend even more on.
- Easier to think about and link parent sample IDs or related sample IDs (i.e. if two samples have an "adjacent" relationship that we want to make explicit). In file annotation, users can't really do this. 

### Relevant examples
- For SRA (but not for GEO), NCBI does require submitting [biosample info first](https://www.ncbi.nlm.nih.gov/biosample/docs/submission/faq/) and NCBI ids are minted.

### Challenges
Unlike publications, where there is effectively a global ID authority (PUBMED), we don't have such a possibility for a primary key for samples. But there are many local ID authorities such as a biobank and each lab could be considered the local authority (but if applicable, the biobank is still the better local authority).

Currently, unless there is a clear biobank source that allows linkage, we make the assumption that a study sourced their samples from a _newly assembled cohort_ where the patients have not taken part in other studies. 

From experience, users might have trouble resolving all the ID-related properties that could be in this table...
- `individualID`
- `specimenID`
-  `aliquotID`  
- `parentSpecimenID`
- `individualIDSource` -- New property to refer to where the primary source of the donors, and where the `individualID` came from. Hopefully it is a known `biobankName` value, but the source can be the lab name in common cases as described above.
- `specimenIDSource` -- This currently exists but is a little unclear. Some studies might use it to actually mean `individualIDSource`, others might use it to make clearer where `specimenID` came from (?). Current usage needs to be resolved. 
[...]

PR #395, opened 2024-02-02
Reconceptualized biosample data collection

anngvu:
Initial work on #334. Cross-check with https://dash.harvard.edu/bitstream/handle/1/32725809/Neurology-2016-Hanemann-S40-8.pdf?sequence=1&isAllowed=y.

These templates require a lot of changes in workflows that want to adopt them. I think switching over wholesale for NF is going to be difficult to manage because it's very different from HTAN/Gray Foundation, so I think the original idea was to **_selectively offer/pilot this for just 1-2 projects that might actually benefit_**. 

**This is a prototype (not production use yet) that we merge but need to followup with additional work of setup + testing and understand where to refine things as needed:**

- Set up of a folder called "Biosample" or "Clinical Data" so data could be stored there.
- Set up and test implementation of cross-validation/linkage.
- Figure out how to extract clinical data so somehow so they still show up in the portal.

If above goes well, we'll actually:
- Advertise this new feature to the community and funders such as NTAP.
- Select and discuss rollout for projects that are a good fit for this.
- Add docs/educational materials for contributors on how to use this new and workflow.

 The preview links to review are mainly: 
- HumanIndividualTemplate
- AnimalIndividualTemplate

So if you think these aren't great names for the templates, I [based them on ADKP](https://github.com/Sage-Bionetworks/data_curator_config/blob/196cb2984f11755c339aa42f12ee93ad8c21205e/ADKP/dca-template-config.json#L125) (though for Gray Foundation, HumanIndividualTemplate ~ CohortCoreTemplate, which I think sounds better). We can discuss the best template naming as well.

Update:
@cconrad8 is primary reviewer.


cconrad8:
Nice! it looks like Manifestation Enum was already used here? so the drop downs in the google sheet are showing up with the manifestations rather than what was intended? 
image
  https://github.com/nf-osi/nf-metadata-dictionary/blob/494fe54ffb2e89c360d010f4296006b5fbf80ae5/modules/DCC/Portal.yaml#L63

cconrad8:
And just as a note here that in Gray Foundation the structure is 
'Data' > 'Patient Cohort Data'> 'synapse_storage_manifest_cohortcoretemplate.csv'
in case we want to do something similar

modules/Sample/Diagnosis.yaml
```diff
@@ -5,17 +5,184 @@ enums:
...
+  TumorTreatmentEnum:
+    permissible_values:
+      no specific therapy:
+      chemotherapy:
+      surgery:
+      radiation:
+      targeted therapy:
+      clinical trial:
+  
+  ManifestationEnum:
```
cconrad8, line 85: change to PresenceStatusEnum or something similar?

cconrad8 (changes requested):

allaway (approved):
We discussed this in our 1-1. 


Key takeaways:

-this is great!
-let's consider pilot this manifest in the Sarin project and in Batch 5 of the NF1 biobank project (cc @jaybee84 )
-let's take a look at the 2023 Sundby paper for any updates that we could incorporate
-let's share with NTAP to see if we can align/adopt parts of their clinical data model 
-let's get feedback from Sundby and Rhodes on the NF1 CDM slack
-consider changing name of manifest as described in PR text (prefer Cohort over HumanIndividual) 
-this is good to merge as-is and then we can update after feedback

Issue #398, opened 2024-02-22
Update registered JSON schemas such as superdataset, etc. more automatically


anngvu:
Linked discussion [in Slack](https://sagebionetworks.slack.com/archives/C6KQHAGJX/p1707929919982509).

So this has two parts:
- First, when something like `modules/Assay` changes, the JSON schemas need to be rebuilt, similar to how NF.jsonld is rebuilt. NF.jsonld format "targets" the DCA application, while json format "targets" the Synapse platform.
- If changes are merged, JSON schema needs to be re-registered with Synapse registry.

cconrad8:
@anngvu Can you explain why we have the 

`Validation Schema = superdataset`

 
 in synapse in addition to our data model? 

I’m getting confused on what that part does that our data model doesn’t?

PR #399, opened 2024-02-22
Feat/json schema workflow

anngvu:
This implements needed changes described in #398.
Previously schemas were re-registered manually after updates, but it would be better to build it into the workflow.

> 
> So this has two parts:
> 
>     First, when something like modules/Assay changes, the JSON schemas need to be rebuilt, similar to how NF.jsonld is rebuilt. NF.jsonld format "targets" the DCA application, while json format "targets" the Synapse platform.

You can see this in how files changed not only includes just `.jsonld` but also the `.json`s now.

>     If changes are merged, JSON schema needs to be re-registered with Synapse registry.

This should mostly work -- it is in the `propagate.yaml` workflow. It is just sending a request to the Synapse server with the new schemas.

Update: adding @cconrad8 mainly for awareness bc asked about this last sprint.


cconrad8 (approved):

cconrad8:
Thanks! I will comment on the issue with just one follow-up to help me understand but unrelated to the PR

PR #400, opened 2024-02-29
Update Genotype.yaml

cconrad8:
It appears the ' is showing up in the manifests - trying to see if removing this here works


cconrad8:
Issue here: image

Changed to " to try that instead

cconrad8:
image
It might be irrelevant because it doesn't show up in sheets once it's clicked

anngvu:
I think it's always been like this, actually, because it's GoogleSheets knows "this is a string, not the start of a formula".

PR #401, opened 2024-03-07
Fix case-insensitive bug from schematic upgrade

anngvu:
@cconrad8 encountered this during mtDCA testing and I just realized this.

So before we upgraded to schematic 24.1.1 in https://github.com/nf-osi/nf-metadata-dictionary/pull/396, enums were capitalization-sensitive, so you could use `DataSubtype` as range and it would be fine because schematic still distinguished the _property_ `dataSubtype` from the _valid values_ `DataSubtype`. But now anything _valid values_ really needs to be DataSubtype**Enum** now.

Can see this by comparing templates generated in [PR 396](https://github.com/nf-osi/nf-metadata-dictionary/pull/396) with e.g. ones *_before_*, such as  [PR 393](https://github.com/nf-osi/nf-metadata-dictionary/pull/393).

### DIFF
--- a/NF.jsonld
+++ b/NF.jsonld
@@ -9,6 +9,3688 @@
   },
   "@id" : "http://schema.biothings.io/",
   "@graph" : [ {
+    "@id" : "bts:BodyPartEnum",
+    "@type" : "rdfs:Class",
+    "rdfs:comment" : "TBD",
+    "rdfs:label" : "BodyPartEnum",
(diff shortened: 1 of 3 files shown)


cconrad8:
Could you show me a screen shot of where you mean? Having a hard time seeing! Although I believe you!

image

image

anngvu:
@cconrad8 It is pretty subtle, that's probably why we missed it! 
annotely_image

PR #408, opened 2024-03-14
Feat/meth array template

### LINKED ISSUE #405: Methylation array assay template
The base EpigeneticsAssayTemplate really means something sequencing-based because it inherits from `BulkSequencingAssayTemplate`. However, DNA methylation arrays are still very common and it would be better to have a more specific and suitable template, because some attributes required for sequencing data wouldn't make sense. For example, [ADKP model](https://raw.githubusercontent.com/adknowledgeportal/data-models/main/AD.model.jsonld) has `AssayMethylationArrayMetadataTemplate`.

anngvu:
Close #405. Basically, the current EpigeneticsAssayTemplate assumes bisulfite sequencing, but we do need to add a template for methylation array data for attributes to make more sense, such as having channels, etc. ADKP model has such a template, but we're adapting Genomics Data Common's https://docs.gdc.cancer.gov/Data_Dictionary/viewer/#?view=table-definition-view&id=raw_methylation_array because it is a better standard.

Update: here is an example with metadata filled in (translated from GEO) for [Harish Vasudevan dataset](https://docs.google.com/spreadsheets/d/1LANNEC7dSVTXH6NVNUb2MFqEmochwf82nO38iIKADzw/edit#gid=0).


NF.jsonld
```diff
@@ -7675,6 +7781,28 @@
...
+    "rdfs:comment" : "User-specified identifier for the specific well of the plate used to prepare the sample for analysis.",
+    "rdfs:label" : "plateWell",
+    "rdfs:subClassOf" : [ ],
+    "schema:isPartOf" : {
+      "@id" : "http://schema.biothings.io/"
+    },
+    "sms:displayName" : "plateWell",
+    "sms:required" : "sms:false"
+  }, {
+    "@id" : "bts:chipID",
```
allaway, line 7796: @anngvu - isn't this the same as `platform`? Could we consolidate with that concept (and perhaps also we could specify GEO platform ids for arrays....? :) (

allaway (approved):
LGTM. I made one optional suggestion regarding chipID

anngvu:
Thanks @allaway -- GDC actually already had `platform` as a separate field with the expected enums, so I think chipID is an optional user-specified field intended for batch management.

PR #414, opened 2024-04-01
Update Parameter.yaml

anngvu:
https://sagebionetworks.jira.com/browse/NFOSI-774

Edit: tests are failing due to `ImportError: cannot import name 'TypeAliasType' from 'typing_extensions'`, and we haven't changed any versions or testing code since last run https://github.com/nf-osi/nf-metadata-dictionary/pull/412, so I think this may be an Actions environment issue that may resolve itself or need to be resolved separately.


allaway:
Interesting, I am also getting this error when I try to run schematic CLI....

PR #415, opened 2024-04-02
add specimenID to bts:ProcessedAlignedReadsTemplate

allaway:
SpecimenID is useful here as it often contains info about different replicates, treatments, etc, in cell line sequencing

### DIFF
--- a/NF.jsonld
+++ b/NF.jsonld
@@ -7745,6 +7745,8 @@
       "@id" : "bts:workflowLink"
     }, {
       "@id" : "bts:auxiliaryAsset"
+    }, {
+      "@id" : "bts:specimenID"
     } ],
     "sms:displayName" : "ProcessedAlignedReadsTemplate"
   }, {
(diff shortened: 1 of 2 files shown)


allaway:
Hmm, not sure if I broke something or the tests are broken..

`ImportError: cannot import name 'TypeAliasType' from 'typing_extensions' (/opt/hostedtoolcache/Python/3.10.12/x64/lib/python3.10/site-packages/typing_extensions.py)`

anngvu (approved):
Thanks -- you can ignore the test results, there's suddenly some dependency issue on GH Actions

anngvu (changes requested):
Actually, I just realized there may be another fix needed, since this inherits from BiologicalAssayDataTemplate, which has `specimenID` -- let me make suggestion

anngvu (approved):
OK, from my comments I see why it wasn't at that level originally -- all good!

allaway:
> OK, from my comments I see why it wasn't at that level originally -- all good!
I went thru the exact same thought process! :) I am happy to move it up a level if that makes sense more generally. I can imagine other assays might benefit from specimenID info.

PR #427, opened 2024-04-17
Update props.yaml

cconrad8:
to correct the warning that is given when age is blank.  

adding '- required ' should ignore the validation rule of number if there is nothing entered

### DIFF
--- a/NF.jsonld
+++ b/NF.jsonld
@@ -9,3777 +9,6 @@
   },
   "@id" : "http://schema.biothings.io/",
   "@graph" : [ {
-    "@id" : "bts:BodyPartEnum",
-    "@type" : "rdfs:Class",
-    "rdfs:comment" : "TBD",
-    "rdfs:label" : "BodyPartEnum",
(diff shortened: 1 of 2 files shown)


cconrad8:
Hmmm @anngvu how do I add validation rules properly?

cconrad8:
In Gray Foundation it looks like we use something like this, but here I used `pattern:` which i found on linkml docs . Thoughts on what the correct syntax is?

   ```
 annotations:
      validationRules: regex match ^\d*\.?\d*$|^unknown$
```

anngvu:
@cconrad8 

- For validation rules in general: Use schematic syntax documented in Confluence -- it does make things a lot easier to remember and consistent with all schematic requirements. I'll add a note about that in the README.
- More specifically in this case: I don't think it makes sense that schematic is outputting warnings that "age is not a num" when age is not required and the age field is blank... so it shouldn't be fixed by us changing validation to regex, it should be fixed in schematic.

anngvu (commented):
Added explanation why not recommended

PR #430, opened 2024-04-19
Update for individualID

anngvu:
Related to [FDS-1968](https://sagebionetworks.jira.com/browse/FDS-1968).

So we should allow a list of IDs anyway, even aside from the above issue where I think schematic gets confused during submission when a field that's not supposed be a list ends up having commas (?).

### DIFF
--- a/NF.jsonld
+++ b/NF.jsonld
@@ -4592,16 +4592,17 @@
     "sms:displayName" : "batchID",
     "sms:required" : "sms:false"
   }, {
-    "@id" : "bts:individualID",
-    "@type" : "rdfs:Class",
-    "rdfs:comment" : "A unique identifier (non-PII) that represents the individual from which the data came. This could be a patient or animal ID.",
-    "rdfs:label" : "individualID",
     "rdfs:subClassOf" : [ ],
+    "@id" : "bts:individualID",
     "schema:isPartOf" : {
       "@id" : "http://schema.biothings.io/"
     },
-    "sms:displayName" : "individualID",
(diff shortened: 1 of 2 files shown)


allaway (approved):
LGTM! One comment, we need to update the individualID attribute here to a stringList: 
https://www.synapse.org/#!Synapse:syn16858331/tables/

Will we need to update all of the schemas for the project fileviews too?

anngvu:
@allaway -- thanks, good reminder -- created ticket [here](https://sagebionetworks.jira.com/browse/NFINT-945); it does appear there needs to be a bit of cleanup for a couple of instances where ids were submitted as single long strings (e.g. "patient1, patient2, ...") before we can convert, because I'm getting warnings about table column sizes.

cconrad8:
Nice @anngvu ! I'm just curious where do these validation rules come from? are they specific to linkml or jsonld schema? wondering where to find "list like" and other validation rules

allaway:
@anngvu , thanks for creating the ticket! 

I am pretty sure [this function](https://github.com/nf-osi/nfportalutils/blob/4320d35b99b48466de41c5337e0ff6dd8da5b2f4/R/convert_to_stringlist.R#L2) will do what you want, though I'm not sure if it still works, since it has been a while!

cconrad8:
nvm! @mialy-defelice sent me to the page https://sagebionetworks.jira.com/wiki/spaces/SCHEM/pages/3302785036/Schematic+Validation

Issue #428, opened 2024-04-17
Add RCC to fileFormat - and other comments to support NanoString-based spatial transcriptomics


allaway:
I got rid of the template because I don't have time to research/scope what needs to be done here, but I realized our data model doesn't currently support Nanostring spatial transcriptome data, like this: https://0-www-ncbi-nlm-nih-gov.brum.beds.ac.uk/geo/query/acc.cgi?acc=GSE239561 which is stored in RCC format.

allaway:
After some digging and conversation with @adamjtaylor I realized I was just _assuming_ this was spatial data - this is actually nCounter data.

PR #432, opened 2024-04-22
add geomx template - draft

allaway:
This PR is to address #428 . 

-I created a GeoMX specific template because it doesn't need a lot of the sequencing related terms, as it is a probe-based transcription assay. I tried to map from HTAN's data model for this, though it seems like most of the details are captured in libraryPrep and libraryPreparationMethod (just more granular/more attributes in HTAN model). 

-I didn't include a value (e.g. `Synapse ID of GeoMx DSP PKC File,`) it doesn't appear to be something that is available on GEO. I'm not sure if this is a nice to have or required for data re-use. We should ask the data contributors.

-I also added RCC file format and did some general meaning -> source cleanup in the FileFormat component. 

Remaining questions that should be addressed before merging: the HTAN template includes four values for the assay that are possible (`GeoMx DSP Assay Type,The assay type which was used for the GeoMx DSP pipeline.,"RNA nCounter, Protein nCounter, Protein NGS, RNA NGS",,,TRUE,Spatial Transcriptomics,,,`). Currently the Assay is just "spatial transcriptomics". Should we deprecate this and add more detailed terms? Or, alternatively, is this more accurately described or captured  by`platform` metadata with the way we currently scope this term?


allaway:
Also requesting feedback from @adamjtaylor if you have any. I hope you don't mind me butchering/retrofitting the HTAN data model to fit ours. :)

allaway:
This seems to indicate the PKC file is required. https://nanostring.com/products/geomx-digital-spatial-profiler/geomx-dsp-configuration-files/

We could just archive all of them on Synapse.... Maybe HTAN has already done this.

allaway:
Closing b/c I misunderstood the data and need to make some changes!

PR #434, opened 2024-04-26
Patch main-ci workflow

anngvu:
### What this does

- Update to schematic version 24.4.1, which supposedly should be faster for template generation and will appear soon in DCA
image

- Better recognition of when to commit, easy fix for below instances (which happens 2-3% of the time?)
image

  - Additional explanation: some changes, like edits to `notes:` and `source:` in the src files in `modules` are like silent mutations and do not lead to meaningful changes in the jsonld, json files

- Remove the `-rr` flag to _exactly match our DCA config_ https://github.com/Sage-Bionetworks/data_curator_config/blob/prod/NF-OSI/dca_config.json#L25

#### Update 1
@allaway, it looks like generation works fine, something has changed with validation (compare with the tests from your recent PR https://github.com/nf-osi/nf-metadata-dictionary/pull/433 which used 24.2.1). 

Error:
image

Looked a bit at the code around the indicated line... I will modify the mock manifests to understand what the expected behavior now when we specify a "num" rule.
https://github.com/Sage-Bionetworks/schematic/blob/c51af2d8419645fa04a13cc387e7811f1afe5b35/schematic/models/validate_attribute.py#L1025

#### Update 2

OK, I realized we need to remove the `-rr` flag now since we no longer want to restrict Great Expectations rules. Now the only issue is
```
error: '' is not of type 'array'
[['2', 'modelSystemName', "'' is not of type 'array'", '']
```


anngvu:
Hey @allaway @cconrad8, I think we can merge this, just wanted you both to be aware that:
- Some validation tests are failing in 24.4.1, so just expect to see this on future PRs for now and not be too concerned -- have communicated this issue to Amy during D&T meeting, so we just wait to see what they say/if there's a fix -- we might need a temp workaround, e.g. by modifying a validation rule, but that would be a separate PR anyway
- Also, this should fix the more minor issue with empty commits

allaway (approved):
this LGTM, thank you @anngvu !

PR #443, opened 2024-05-07
add GE MRI

allaway:
For NFOSI-810

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -10,10 +10,12 @@ on:
   
   workflow_dispatch:
 
+env:
+  SCHEMATIC_VERSION: 24.5.1
   
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


allaway:
Converting this back to draft b/c I missed some things...

allaway:
OK, this is actually ready for review now 😓

anngvu (commented):
Summary of additional discussion: resolution for now is to keep as single value to retain dropdown with a special selection string value like "Disease1,Disease2,Disease3" until we have better UI.

allaway:
> Summary of additional discussion: resolution for now is to keep as single value to retain dropdown with a special selection string value like "Disease1,Disease2,Disease3" until we have better UI.

Actually, I was thinking about just pushing the value "Disease1,Disease2,Disease3" "manually", ie w/o validation. But I could also add it as a value to the schema. I believe these contributors will not be adding more data to their project, which makes me lean towards the skip-validation approach (it's 3 files total). What do you think?

allaway:
Actually, now that I think about it, if I'm skipping validation, should I just annotate it as a stringList anyway?

anngvu:
@allaway Yeah, let's just annotate it as stringList for these 3 files then. We can skip adding the unwieldy "Disease1,Disease2,Disease3" since there's already human validation, technically.

anngvu:
Superseded by #457

PR #451, opened 2024-05-14
Patch/schematic 24.5.1

### LINKED ISSUE #449: Update schematic version, add'l conversion, and CLI revision
- 24.5.1 should resolve the `list like` issue.
- Add `schematic schema convert` on JSON-LD
- Remove `data_model_labels` flag

anngvu:
Close #449

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -10,6 +10,8 @@ on:
   
   workflow_dispatch:
 
+env:
+  SCHEMATIC_VERSION: 24.5.1
   
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


allaway:
@anngvu , thanks for doing this! I had a question about the superdataset/dataset assay values disappearing. is that expected?

anngvu:
@allaway Actually, good catch. The JSON Schema build was trying to select ids that were changed, so just had to make sure that the build ran in the right order. But this reminds me I'll also have to change any other scripts that reference assay and institution values by using their key ids in the model.

PR #453, opened 2024-05-24
Update PortalDataset.yaml

anngvu:
Here's a mockup of what this might look like in the UI (maybe fancier than a link, should be an icon specific to the viewer). This would be useful for known viewers that can be linked to via a constant accession/url, such as cBioPortal, GEO, CELLxGENE.

Update: Suggestion for a better term other than "externalViewer" welcome.

image

### DIFF
--- a/modules/Template/PortalDataset.yaml
+++ b/modules/Template/PortalDataset.yaml
@@ -72,7 +72,7 @@ classes:
       series:
         range: string
         required: false
-      externalViewer:
+      visualizeDataOn:
         range: string
         multivalued: true
         required: false
--- a/modules/Template/PortalDataset.yaml
+++ b/modules/Template/PortalDataset.yaml
@@ -72,6 +72,10 @@ classes:
       series:
         range: string
         required: false
+      externalViewer:
+        range: string
+        multivalued: true
+        required: false
       yearProcessed:
         range: integer
         required: false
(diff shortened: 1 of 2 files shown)


anngvu:
@allaway Yeah, I know there's the attribute-per-viewer option, and I have reasons for not liking it. But anyway, there wasn't a pros and cons written down anywhere that I can find, so created issue https://github.com/nf-osi/nf-metadata-dictionary/issues/455 for more formal discussion/design documentation.

Good to know about the identifiers prefix!

jaybee84:
Some alternate options for `external viewer` : 
1. `visualize data on` -> does not work for link to GEO/SRA (that might be okay since this is specifically for viewers)
2. `explore data` 
3. `explore insights` -> does not work for link to GEO/SRA (that might be okay since this is specifically for viewers)
4. `additional portal`
5. `More Details`
6. `Additional resources`
7. `Extra Content`
8. `Further Details`

4-8 are more generic and may work for links other than visualization portals.

cconrad8 (approved):

allaway:
@jaybee84 I likke `Explore Insights` and `visualize data on`! If we start linking to other _raw data_ I think we should categorize that separately anyway, so probably not too much of a problem (does not work for link to GEO/SRA (that might be okay since this is specifically for viewers)

modules/Template/PortalDataset.yaml
```diff
@@ -72,6 +72,10 @@ classes:
       series:
         range: string
         required: false
+      externalViewer:
```
allaway, line 75:
```suggestion
      visualizeDataOn:
```

Suggesting from @jaybee84

PR #458, opened 2024-05-30
Update devcontainer

anngvu:
OK, I see what's going on. So there was a `typing_extensions` issue that cropped up around April, and a workaround/fix was put into https://github.com/nf-osi/nf-metadata-dictionary/pull/421 but I neglected the .devcontainer 😞 .

Before patch:
image

After patch:
image

### DIFF
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
-pip3 install schematicpy==24.1.1
+pip3 install schematicpy==24.5.1 ipython==8.18.1
 sudo bash < <(curl -s https://raw.githubusercontent.com/babashka/babashka/master/install)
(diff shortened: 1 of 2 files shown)


.devcontainer/post-create.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
-pip3 install schematicpy==24.1.1
+pip3 install schematicpy==24.5.1 ipython==8.18.1
```
linglp, line 3: Will it be better to do: `pip3 install --upgrade schematicpy`, to ensure that the latest version of schematicpy gets installed?

allaway: LGTM! @linglp , I think we have had better luck pinning the version for our CI workflows, new versions have sometimes had breaking changes that we didn't expect (until the workflow breaks :) )

allaway (approved):

PR #476, opened 2024-07-12
Add new single cell assays

allaway:
For https://github.com/nf-osi/dcc-site/pull/1107/files


anngvu (commented):
Hmm, we should probably make hyphen usage more consistent across the single cell labels later

PR #496, opened 2024-08-19
Update PortalDataset.yaml

cconrad8:


cconrad8:
added isGlobalFeatured to indicate which datasets should be listed in the data catalog

anngvu (approved):
Merged in changes upstream, which included a fixed version for LinkML to resolve some weird syntax bug, reran workflow, we should be all good now!

Issue #492, opened 2024-08-13
Additions from new data sharing plans or annotation help requests


anngvu:
Running list:
- cell painting - https://www.nature.com/articles/nprot.2016.105
- ZetaView (platform) - https://sagebionetworks.jira.com/jira/servicedesk/projects/NFOSI/queues/custom/7/NFOSI-1086

Issue #486, opened 2024-07-26
add library kit as required term for scRNAseq manifest


allaway:
a recent user asked for this information, which was missing from the file annotations: https://sagebionetworks.jira.com/browse/NFOSI-853

allaway:
resolved by #500

PR #500, opened 2024-08-19
Patch/issues 486 492

anngvu:
Issues #486 + #492

@allaway We're going to use libraryID to capture library kit, but let me know if it makes sense to use something else.


allaway:
This makes sense, we'll just need to port over the annotations currently under libraryID in the portal. Food for thought: if we can dig up some controlled terminology for this, it would be super helpful. The world of probe sets is really painful- lots of versions of the "same" kit and there can be different probeset bed files associated with each of these. 

Here;s an example list that UCSC maintains: https://genome.ucsc.edu/cgi-bin/hgTrackUi?db=hg38&g=exomeProbesets

PR #502, opened 2024-08-20
First attempt at adding props to RNAseqTemplate

aditya-nath-sage:
Added requested fields to Data.yml in RNASeqTemplate. 

Do I need to update any testing resources? Was this the correct Assay to update?

### DIFF
--- a/modules/Template/Data.yaml
+++ b/modules/Template/Data.yaml
@@ -172,6 +172,10 @@ classes:
     is_a: BulkSequencingAssayTemplate
     description: Template for describing raw data from (bulk) RNA-sequencing
     slots:
+    - genePerturbed
+    - genePerturbationType
+    - genePerturbationTechnology
+    - experimentalCondition
     annotations:
       requiresComponent: ''
       required: false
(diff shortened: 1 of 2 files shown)


modules/Template/Data.yaml
```diff
@@ -178,6 +178,10 @@ classes:
         assay:
         - RNA-seq
         - lncRNA-seq
+        - genePerturbed
```
anngvu, line 181: Hey Aditya, this is the correct template, but in our walkthrough, a similar update was made by adding attributes to under `slots:`. Remember that this assay annotation is meant for mapping, and not actually used.

anngvu (changes requested):
See comment

modules/Template/Data.yaml
aditya-nath-sage: Ahh got it thank you!

anngvu (commented):
(Also, might want to reset and use more recent updates from main to avoid the conflict.)

anngvu (approved):
Good job @aditya-nath-sage! Note: the two failing tests are expected due to some schematic issues. Merging.

Issue #479, opened 2024-07-16
Improve portal "publication" concept to differentiate between primary and secondary uses of studies


anngvu:
From https://sagebionetworks.jira.com/browse/NFINT-805. Depends on #478.

> Recently, we have been asked a few times about what publications are the “primary” publication for a dataset and what publications represent secondary use of data. It would be great to differentiate these in the portal data and UI, kind of like on the NF Tools Database. 
> 
> However, there are a couple of questions that come to mind for me:
> 
> 1 - What is a secondary publication? if a PI re-uses their own data for the same study, should we call that primary or secondary? My vote would be secondary. 
> 
> 2 - Can a study have multiple primary publications (yes, probably, because it is not 1-1 datasets to studies, but happy to discuss this more.

allaway:
Lol, I was about to vote for these questions, but then I realized I posted the initial issue 😅

PR #542, opened 2024-11-12
Created prelim yaml file for PortalPublication

aditya-nath-sage:
Created a quick yaml file for PortalPublication


anngvu (changes requested):
To summarize: We needed to formalize the NF publication model for awhile now. It still needs to be backwards-compatible with the current model used for the portal at https://www.synapse.org/Synapse:syn16857542/tables/ while adding useful concepts and aligning with other projects if possible. 

For example, some concepts we find especially useful from MC2 are `grantNumber` and `accessibility`; these are additions into the new formal model. Another addition needed is to address #479 (originally raised by @allaway) -- maybe add something called `type` with possible values "primary" and "secondary"? 

Changes needed for this PR:
- Implement minor naming convention and removal changes
- Implement addition to address #479 
- Merge the Makefile update
- Test the compilation workflow to regenerate JSON with changes
- Add a new make step to https://github.com/nf-osi/nf-metadata-dictionary/blob/b67754a7397d4eb04c81f0824ba66bec8b6baef7/.github/workflows/main-ci.yml#L38

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+        range: string
+        required: true
+      year:
+        range: integer
+        required: true
+      diseaseFocus:
+        range: string
+        multivalued: true
+        required: false
+      authors:
```
anngvu, line 46: Change to `author` for backwards-compatibility

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,71 @@
...
+      publicationDoi:
+        range: string
+        required: true
+      publicationJournal:
+        range: string
+        required: true
+      pubmedId:
+        range: integer
+        required: true
+      pubmedUrl:
```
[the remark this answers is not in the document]
anngvu: Still need to remove `pubmedUrl`

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+    is_a: Template
+    description: >
+      A generalized data model template for Publications to be hopefully used across dccs
+    notes:
+    - This structure is based on a CSV file from MC2 annotations with customized fields as attributes.
+    attributes:
+      grantNumber:
+        range: string
+        required: true
+      consortiumName:
```
anngvu, line 18: Remove `consortiumName` as it shouldn't be the base class -- not everyone will use it; we certainly don't

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+        range: string
+        required: true
+      themeName:
```
anngvu, line 21: The base needs to be minimal and streamlined -- remove `themeName`

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+      pubmedUrl:
+        range: string
+        required: true
+      title:
+        range: string
+        required: true
+      year:
+        range: integer
+        required: true
+      diseaseFocus:
```
anngvu, line 42: `diseaseFocus` should be moved to NF subclass.

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+        multivalued: true
+        required: false
+      authors:
+        range: string
+        multivalued: true
+        required: false
+      abstract:
+        range: string
+        required: true
+      assay:
```
anngvu, line 53: Remove `assay`

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+        range: string
+        multivalued: true
+        required: false
+      accessibility:
+        range: string
+        required: true
+
+  NFPublication:
```
anngvu, line 61: Can you change name of class to `PortalPublication` to match the rest of the portal entities (`PortalStudy`, etc.)?

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+    is_a: PublicationTemplate
+    description: A slightly more specialized publication concept intended for the specific scope of the NF Portal.
+    attributes:
+      studyUse:
```
anngvu, line 65: I think instead of leaving this as string, we create an enum called "StudyProductTypeEnum" (or something else more fitting if you can think of it) with values "primary" and "secondary". Use the example enum syntax for Sex, in https://github.com/nf-osi/nf-metadata-dictionary/blob/main/modules/Sample/Sex.yaml. For now, just put the new enum as a file in https://github.com/nf-osi/nf-metadata-dictionary/tree/main/modules/Other. Then we can reuse this enum later on if we need.

anngvu, line 65: Can you also add a description to this one, since `studyUse` is not totally obvious, unlike things like `title`?

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,70 @@
...
+        range: string
+        required: true
+      nfType:
```
anngvu, line 68: What is `nfType`?

anngvu (changes requested):
Using the inheritance syntax looks good, but I've reiterated some suggestions for items that weren't really resolved, and requested additional changes for what should be in the base class vs the subclass, as well as refining more bespoke and trickier attributes.

modules/Template/PortalPublication.yaml
anngvu: Also, this is one of those tricky concepts where it's more debatable which labels offer the best fit, so feel free to generate more ideas with LLM and poll @cconrad8 and @allaway.

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,71 @@
...
+      publicationThemeName:
+        range: string
+        required: true
+      publicationDoi:
+        range: string
+        required: true
+      publicationJournal:
+        range: string
+        required: true
+      pubmedId:
```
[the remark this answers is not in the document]
allaway: Another note here, I think PMID should be a string, not an integer. The only reason for this is that when we preface the PMID with an ID, the portals/synapse will automatically convert it to a pubmed link: https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/9e4da441b2ab066bd2180ac071877173fe9573e5/packages/synapse-react-client/src/components/GenericCard/Linkify.tsx#L34

modules/Template/PortalPublication.yaml
aditya-nath-sage: I was thinking this would be the disease type of nf (Type 1, 2, schwannomatosis)

modules/Template/PortalPublication.yaml
aditya-nath-sage: This is probably the NFType attribute I made, will replace with diseasefocus

modules/Template/PortalPublication.yaml
aditya-nath-sage: Moved to the NF specific class

modules/Template/PortalPublication.yaml
aditya-nath-sage: Replaced with diseasefocus

modules/Template/PortalPublication.yaml
aditya-nath-sage: Definitely prefer to make it an enum!

cconrad8: Hmm trying to think! Can you clarify @aditya-nath-sage what does "studyUse" mean in this context?

aditya-nath-sage: Definitely, studyUse is related to issue [#479](https://github.com/nf-osi/nf-metadata-dictionary/issues/479)

It relates to whether the publication is a primary or secondary use of data made by the PI who wrote the publication.

Questions (from the issue):

1 - What is a secondary publication? if a PI re-uses their own data for the same study, should we call that primary or secondary? My vote would be secondary.

2 - Can a study have multiple primary publications (yes, probably, because it is not 1-1 datasets to studies, but happy to discuss this more.

modules/Other/PublicationEnum.yaml
```diff
@@ -0,0 +1,9 @@
+enums:
+  StudyTypeEnum:
+    permissible_values:
+      Primary:
+        description: This publication is used as a primary publication for the dataset
+        meaning: 
+      Secondary:
+        description: This publication is used as a secondary publication for the dataset
+        meaning:
```
anngvu, line 9: Linting issue -- add a line ending at the end of file or make script won't work

modules/Other/PublicationEnum.yaml
```diff
@@ -0,0 +1,9 @@
+enums:
+  StudyTypeEnum:
```
anngvu, line 2: Enum name needs to match what's defined in `range` (PortalPublication.yaml line 47). Let's call this `PublicationTypeEnum` and update `range` there accordingly.

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,60 @@
...
+      accessibility:
+        range: string
+        required: true
+
+  PortalPublication:
+    is_a: PublicationTemplate
+    description: A slightly more specialized publication concept intended for the specific scope of the NF Portal.
+    attributes:
+      studyUse:
+        range: PublicationEnum
```
anngvu, line 47: The range value doesn't match the new enum. See comment in modules/Other/PublicationEnum.yaml

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,60 @@
...
+        required: true
+      diseaseFocus:
+        range: string
```
anngvu, line 50: We do have a `DiseaseFocusEnum` (see PortalStudy for example), so we can stick it in the range here to reuse it

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,60 @@
...
+        required: true
+      accessibility:
+        range: string
+        required: true
+
+  PortalPublication:
+    is_a: PublicationTemplate
+    description: A slightly more specialized publication concept intended for the specific scope of the NF Portal.
+    attributes:
```
anngvu, line 45: Missing a `fundingAgency` attribute (this can be multi-valued)

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,60 @@
...
+        required: true
+      author:
+        range: string
+        multivalued: true
+        required: false
+      abstract:
+        range: string
+        required: true
+      accessibility:
+        range: string
```
anngvu, line 39: Can you define enum for accessibility as well? ("Open Access", ...)

modules/Template/PortalPublication.yaml
```diff
@@ -0,0 +1,60 @@
...
+    description: A slightly more specialized publication concept intended for the specific scope of the NF Portal.
+    attributes:
+      studyUse:
+        range: PublicationEnum
+        required: true
+      diseaseFocus:
+        range: string
+        multivalued: true
+        required: false
+      assay:
```
anngvu, line 53: Let's remove `assay` for now since we don't curate assays from publications.

anngvu (changes requested):
Requested changes. You can make changes and then run `make PortalPublication` in GH workspace to test/see the the compiled JSON schema under the registered-json-schemas folder.

aditya-nath-sage:
Made the requested changes but got error 127 when running make PortalDataset

anngvu (approved):
Thanks for making changes and going through the slightly more "advanced" and JSON schema-centric editing. Updated Makefile compilation and a couple of small things to finalize, so we can merge.

Issue #532, opened 2024-10-16
Add template for ChIP-seq assay


allaway:
**Describe the issue with the metadata**
It would be helpful to have a chip-seq specific template. EpigeneticsSequencingAssay has content that is not relevant (e.g. bisulfiteConversionKitID) and is missing relevant content (e.g. assayTarget for the pulldown). 

**Additional comments**

anngvu:
ChIP-seq specific template could also include antibody used. https://pmc.ncbi.nlm.nih.gov/articles/PMC3431496/#sec13

changtotheintothemoon:
Hi Robert @allaway,

I found the section on Epigenetics sequencing assays here:
[https://github.com/nf-osi/nf-metadata-dictionary/blob/main/modules/Template/Data.yaml#L271-L299](https://github.com/nf-osi/nf-metadata-dictionary/blob/main/modules/Template/Data.yaml#L271-L299)

For ChIP-seq, here’s a list of possible types, though I'm not sure how many of these are relevant to NF projects. Please let me know if there are specific types we should focus on implementing:

- **Transcription Factor (TF) ChIP-seq**
- **Polymerase ChIP-seq** (e.g., RNA Pol II ChIP-seq)
- **Cohesin and CTCF ChIP-seq**
- **Mediator Complex ChIP-seq**
- **Histone Variant ChIP-seq** (e.g., H2A.Z, H3.3)
- **Acetylation-Specific or Phosphorylation-Specific ChIP-seq**
- **DNA-Binding Protein ChIP-seq** (e.g., p300/CBP ChIP-seq)
- **Chromatin Remodeler ChIP-seq**

Let me know your thoughts or any suggestions on implementing this. Thanks!

PR #543, opened 2024-11-12
ChIPSeq Template Created

changtotheintothemoon:
Hi @allaway @anngvu

I created a new initial branch for adding a new template for the ChIP-Seq assay:  
Related Issue: [#532](https://github.com/nf-osi/nf-metadata-dictionary/issues/532)

Please review the template and let me know if there are additional sections that need to be added.

@allaway mentioned creating a new section for the antibody in `modules/Sample/SpecimenProcessing.yaml`. I have not modified this yet.  
He suggested using the OLS/NCIT definition for the ChIP-Seq Antibody Name: [NCIT_C177613](https://www.ebi.ac.uk/ols4/ontologies/ncit/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FNCIT_C177613?lang=en)

@anngvu, what are your suggestions regarding adding the antibody to the template?

Thanks,  
James

### DIFF
--- a/modules/Template/Data.yaml
+++ b/modules/Template/Data.yaml
@@ -268,7 +268,7 @@ classes:
     close_mappings:
     - htan:BulkWESLevel1
 
-  EpigenomiscAssayTemplate:
+  EpigenomicsAssayTemplate:
     is_a: BulkSequencingAssayTemplate
     description: Template for describing raw data from epigenetics sequencing assays such as bisulfite sequencing. 
     slots:
@@ -290,8 +290,8 @@ classes:
     - htan:BulkMethylation-seqLevel1
 
   EpigeneticsAssayTemplate:
-    is_a: EpigenomiscAssayTemplate
-    description: Alias for EpigenomiscAssayTemplate for backwards-compatibility.
+    is_a: EpigenomicsAssayTemplate
+    description: Alias for EpigenomicsAssayTemplate for backwards-compatibility.
     slots:
     annotations:
       requiresComponent: ''
--- a/modules/Template/Data.yaml
+++ b/modules/Template/Data.yaml
@@ -303,7 +303,7 @@ classes:
     slots:
       - antibodyID
       - platform
-      - peakCallingAlgorithm # This is not in the modules/props.yaml file. Should we add a new property?
+      - peakCallingAlgorithm
     annotations:
       requiresComponent: ''
       required: false
--- a/modules/Template/Data.yaml
+++ b/modules/Template/Data.yaml
@@ -282,7 +282,6 @@ classes:
         - bisulfite sequencing
         - CAPP-seq
         - CUT&RUN
-        - ChIP-seq
         - ERR bisulfite sequencing
         - HI-C
         - NOMe-seq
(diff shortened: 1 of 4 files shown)


modules/Template/Data.yaml
```diff
@@ -298,6 +297,20 @@ classes:
       requiresComponent: ''
       required: false
 
+  ChIPSeqTemplate:
+    is_a: EpigenomicsAssayTemplate
+    description: Template for describing raw data from ChIP-Seq assays, typically used to identify DNA-protein interactions.
+    slots:
+      - antibodyID
```
anngvu, line 304: You can add a more specific `ChIPseqAntibodyName` property with the NCIT meaning as suggested and use it here. We can either keep it free text or import a list of ChIPseq antibody names, but I'd prefer people use the `antibodyID` as the preferred field anyway -- you can make `antibodyID` required. For additional context, it looks like HTAN just uses single free text `antibodyName` primarily https://github.com/ncihtan/data-models/blob/4b9bb3cfcbb979694479a4a60b7dac71becfbf96/HTAN.model.csv#L226

anngvu (approved):

anngvu (changes requested):
Update props.yaml

modules/Template/Data.yaml
changtotheintothemoon: Is the antibodyID already in free text form?
https://github.com/nf-osi/nf-metadata-dictionary/blob/main/modules/props.yaml#L73-L75

modules/Template/Data.yaml
```diff
@@ -298,6 +297,20 @@ classes:
...
+      - platform
+      - peakCallingAlgorithm # This is not in the modules/props.yaml file. Should we add a new property?
```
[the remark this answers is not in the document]
changtotheintothemoon: L1199-L1214 in props.yaml have been added with list of peakcallingalgorithms

modules/Template/Data.yaml
anngvu: Yes

modules/Template/Data.yaml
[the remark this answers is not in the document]
anngvu: Great!

modules/Template/Data.yaml
changtotheintothemoon: If we make this `antibodyID` slot required, would this also effect other templates that might been using the slot?

changtotheintothemoon (commented):
LGTM

modules/Template/Data.yaml
anngvu: Yes, because the feature to set contextual required is not implemented in schematic yet. In this case, I think it's OK to make it required across all templates that use it -- you would only add antibodyID to the template if you really needed it anyway.

anngvu:
@changtotheintothemoon Once this template is considered "production-ready", to surface this in the app, you should add the template id and display name to the list here: https://github.com/nf-osi/nf-metadata-dictionary/blob/main/dca-template-config.json

changtotheintothemoon (commented):
Approved

anngvu:
@changtotheintothemoon Thanks! I think you're all done except one small issue. So in this review I discovered we have a typo for "Epigenomi**sc**AssayTemplate" 😅  (#550), which is why inheritance is not working in the rendered preview link the tests above. If you fix it, everything should be good for a 2-for-1 deal!

anngvu (approved):

PR #571, opened 2025-01-22
Add LinkML make demo

anngvu:
Hey @nf-osi/dcc-team (if you're interested) and @andrewelamb, for awareness this is how you would use LinkML to generate Excel manifests. I am using the latest version of LinkML. 

We don't use the Excel manifest because we use schematic Googlesheets, but if you ever want to do the comparison, here it is:
- Clone this branch.
- run `make ManifestLinkMLDemo`
- Each manifest is a sheet in the default output `nfosi.xlsx` file (attached: [nfosi.xlsx](https://github.com/user-attachments/files/18497848/nfosi.xlsx))

Here are some findings (not exhaustive):
- Apparently there _is_ an important limitation for valid value dropdowns -- lots of warnings for "total length > 255 characters. Dropdowns may not work properly in /home/avu/sage/nf/nf-metadata-dictionary/nfosi.xlsx"
- Not sure if there's an easy way to get one Excel file per manifest instead of everything in one giant Excel file, as it doesn't seem to be an option to select only the template I want to generate in the `gen-excel` command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ tests/submit/NF.jsonld
 tests/test-suite-report.md
 tests/**/logs
 publish/**
+nfosi.xlsx
 
 **/schematic
 *.schema.json
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,16 @@ analyze:
 convert:
 	bb ./retold/retold as-jsonld --dir modules --out NF.jsonld 
 
+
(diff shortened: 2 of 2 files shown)


andrewelamb:
Thanks @anngvu!
You can do `gen-excel schema.yaml  --split-workbook-by-class` to generate an excel file for each class separately.

PR #585, opened 2025-02-27
Add duo terms + schema.org dataset update

cconrad8:


cconrad8:
@anngvu to push some things forward I'm trying to use NF as a demo for the schema.org + croissant + duo implementation.  Changes include adding the duo terms from https://github.com/Sage-Bionetworks/governanceDUO/blob/main/metadata%20structure.md and updated annotations as much as possible to schema.org terminology to reduce the need for many transformations. This may affect some downstream annotations that we need to go into the portal to fix. 

@aditya-nath-sage for visibility and second reviewer

modules/Template/PortalDataset.yaml
```diff
@@ -20,12 +20,16 @@ classes:
     # - doi (if minted)
     # - createdBy # user id/handle, creator allows setting name more explicitly for citations
     # - datasetItemCount
-    # - datasetSizeInBytes 
+    # - datasetSizeInBytes
     # - publisher # should be autoset
-      title:
-        title: Title
+      name:
```
anngvu, line 25: Originally, title was used to conform to https://www.w3.org/TR/vocab-dcat-3/#Property:resource_title. I think using `name` means we won't have as much control also since in Synapse it's reserved and there are character restrictions on names (so you can't use "/", which can appear a lot). Will there be cases when it might be frustrating to not be able to name your dataset "Data from Phase 1a/b trial" or "Data from NF1+/- model mouse"?  (alternative, we can ask platform if it's possible to do do away with this)

modules/Template/PortalDataset.yaml
```diff
@@ -47,10 +51,22 @@ classes:
         title: License
         range: License
         required: true
-      assay:
-        title: Assay
+      countryOfOrigin:
+        title: Country of Origin
+        range: string
+        required: false
+      measurementTechnique:
```
anngvu, line 58: This is fine, need ticket to migrate `assay` to `measurementTechnique`

modules/Template/PortalDataset.yaml
```diff
@@ -75,8 +91,8 @@ classes:
         range: DiseaseFocusEnum
         multivalued: true
         required: true
-      fundingAgency:
-        title: Funding Agency
+      funder:
```
anngvu, line 94: I've always liked funder more, but I guess we don't have to force consistency

modules/Template/PortalDataset.yaml
anngvu: I guess we can use `alternateName`

anngvu (approved):
Well, I'm not a big fan of breaking changes, but since we don't have a ton of datasets yet, it's easier to change direction earlier than later. Tbh, all projects will need some sort of transformation anyway. Great job adding DUO!

PR #590, opened 2025-03-07
Revisions to improve overall compatibility, clarity, and AI performance

anngvu:
**Update**
For posterity, we had discussion that covered:
-  Let's use DUO human-readable labels after all
- Clarified `subjects` intent (compare and contrast with `keywords) and the scheme we will use (Library of Congress); make it a list allowing multiple values
- Back to `title` to avoid complicating things/clash with Synapse property
- `description` is still annotation
- `creator` should be PI of study, `NF-OSI` can be contributor (for our current processed datasets, NF-OSI should still be creator)


cconrad8:
@anngvu I updated DUO terms to have the full name as the title instead of the code - I think ready to merge this when you're ready

PR #593, opened 2025-03-11
Update script

anngvu:
I don't know how the original script only cares about the first file detected as changed, but this should fix problem mentioned in standup 😅 . You can check the UI on existing bound datasets and it should be updated.


anngvu:
@cconrad8 Tagged for awareness, but I'll go ahead and merge this. Let me know if you see changes not updated anywhere!

PR #607, opened 2025-05-05
Data standardization edamization

### LINKED ISSUE #600: More correctly and fully use EDAM for data type and file format
User story: I would like to take advantage of logic and relationships in EDAM ontology. In order to do that, the meaning of current concepts should be correctly mapped or more concepts potentially imported.

### Data Type Discussion

bio.tools uses EDAM data type concepts and matches data type with operations and software packages, helping users understand what packages can be used to analyze a file or dataset. However, data type is currently very under-developed and not consistently based on EDAM; most concepts were original high-level terms from AD (https://github.com/Sage-Bionetworks/synapseAnnotations), some imports from GDC more recently, and the rest a mix from random ontologies/taxonomies because EDAM itself can lack coverage. 

One main issue for import is the very long list defined by EDAM currently. From my most recent analysis, EDAM defines 800+ data types (see [edam prep script](https://github.com/anngvu/bioc-curation/blob/main/prep-edam.sh) and [extracted results](https://github.com/anngvu/bioc-curation/blob/main/enums.json)), which would be a lot for someone to go through and compare with what we currently have to incorporate EDAM in a lean manner. However, it is more doable now with LLM to help with model development. Some pragmatic expert observations so far: 1) we don't need things like the various IDs, which already removes 500+ terms which otherwise would make the model bloated and confusing for users; 2) we need to reconcile what already exists so we don't import redundant concepts, mapping as closely as possible (use meanings). 

### File Format Discussion

File formats are relatively clean and just need a few meaning conversions to use EDAM for the most part. EDAM provides helpful classification of formats into binary or text formats, etc, and linkage to data type.

anngvu:
Close #600. There are some breaking changes and legacy values we will have to migrate after this.


aditya-nath-sage (approved):

cconrad8 (approved):
looks good to me!

PR #619, opened 2025-05-21
Update Assay.yaml

cconrad8:


anngvu (approved):
Looks good (since it's expected that the aliases will appear only in the LinkML-generated files and not in the schematic JSON-LD which doesn't capture aliases/synonyms). Can ignore the validation test results.

anngvu:
@changtotheintothemoon Not sure if you also need to see this.

moonchangin (approved):
LGTM

changtotheintothemoon (approved):

PR #627, opened 2025-06-13
update meaning/source as needed

allaway:
Digging through docs it's possible this is failing because the PR is from a first-time contributor.


allaway:
closing and going to fix this with cursor :)

Issue #637, opened 2025-06-30
Add automatic json schema to Metadata Dictionary action


cconrad8:
automate the creation of json schemas

Issue #622, opened 2025-06-02
Generate json schemas from linkml


cconrad8:

anngvu:
Consider this done, awesome for getting this new workflow rolling! Left review with some helpful explanations, make sure to merge fixes from #673 to your branch. The next part (#637) should be much simpler and only include two lines in the appropriate step in the workflow file. Can explain more next week.

PR #665, opened 2025-07-14
Batch convert of other LinkML YAML templates -> JSON schema

cconrad8:


cconrad8:
Update: now ready for review!

anngvu (changes requested):
This deletes or replaces some of the Portal schemas we currently have with incorrect version (PortalDataset, PortalStudy, etc.). The issue is probably that these are already in the individual-class format. It also removes a special file "super_rules.json" that we need (update: I think we can move this to a more suitable location and update the current special `make` command that uses it).

~Also, I think it would be better if we move the generated individual-class yaml files to dist/*. They're intermediate representations on the way to the final JSON schemas. Having so many in the Template directory will be confusing.~ Actually, I don't think we technically need these.

anngvu:
Also, it's possible the script can be streamlined by using linkml python package utils... will add example.

gen-json-schema-class.py
```diff
@@ -0,0 +1,225 @@
...
+                return
+            for v in obj.values():
+                resolve(v)
+        elif isinstance(obj, list):
+            for item in obj:
+                resolve(item)
+
+    resolve(props)
+
+    # now normalize, order, and write out as before
```
anngvu, line 169: If use linkml==v1.8.1, don't need this per https://github.com/nf-osi/nf-metadata-dictionary/issues/622#issuecomment-3009674848

gen-json-schema-class.py
```diff
@@ -0,0 +1,225 @@
...
+        continue
+
+    try:
+        raw = json.loads(schema_str)
+    except json.JSONDecodeError as e:
+        print(f"❌ Invalid JSON for {cls}: {e}")
+        continue
+
+    # set ID + title
+    raw["$id"]   = (
```
anngvu, line 133: This would need to be `https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/org.synapse.nf-{classid}` since it would be under nf org

gen-json-schema-class.py
```diff
@@ -0,0 +1,225 @@
...
+
+def slot_enums(sdef):
+    enums = set()
+    if "range" in sdef:
+        enums.add(sdef["range"])
+    if "any_of" in sdef:
+        enums.update(alt["range"] for alt in sdef["any_of"] if "range" in alt)
+    return enums
+
+# ─── STEP 1: GENERATE CLASS-SPECIFIC YAML ─────────────────────────────────────
```
anngvu, line 75: Hmm, so we don't actually need this code for class-specific YAML since LinkML can do this automatically with `gen-json-schema --top-class ..`, and they get kind of overwhelming under src/modules/Template

gen-json-schema-class.py
```diff
@@ -0,0 +1,225 @@
...
+
+# ─── STEP 4: VALIDATE AGAINST SYNAPSE (dryRun) ───────────────────────────────
+print("\n🔨 Validating all dereferenced schemas…")
+syn = synapseclient.Synapse()
+syn.login()
+
+def validate_schema(path: Path):
+    print(f"\n🔍 Validating: {path.name}")
+    data = json.loads(path.read_text())
+    remove_keys(data, {"additionalProperties","metamodel_version","version"})
```
anngvu, line 206: This is great but shouldn't have `remove_keys` here since that was already done in the conversion and wouldn't be validating the schema "as-is"

gen-json-schema-class.py
```diff
@@ -0,0 +1,225 @@
...
+# ─── STEP 2 + 3: JSON SCHEMA GENERATION + DEREFERENCE ────────────────────────
+for tpl in BUILD_DIR.glob("*.yaml"):
+    cls = tpl.stem
+    print(f"\n🔨 Generating JSON Schema for {cls}")
+    schema_str = run_capture([
+        "gen-json-schema",
+        "--top-class", cls,
+        "--inline", "--no-metadata",
+        "--title-from=title",
+        "--closed", str(tpl)
```
anngvu, line 120: We currently used `"--not-closed"` to allow optional annotations.

gen-json-schema-class.py
```diff
@@ -0,0 +1,225 @@
...
+
+    out_json = OUT_DIR / f"{cls}-deref.json"
+    out_json.write_text(json.dumps(ordered, indent=2))
+    print(f"  ✅ wrote {out_json}")
+
+
+# ─── STEP 4: VALIDATE AGAINST SYNAPSE (dryRun) ───────────────────────────────
+print("\n🔨 Validating all dereferenced schemas…")
+syn = synapseclient.Synapse()
+syn.login()
```
anngvu, line 201: Update to use authentication with env var `syn.login(authToken=auth_token)`, since we have token setup in GH workflow.

registered-json-schemas/PortalDataset.json
```diff
@@ -1,60 +1,54 @@
...
-      "description": "Access type for the dataset.",
-      "title": "Access Type",
+      "description": "",
       "enum": [
         "Public Access",
         "Open Access",
         "Controlled Access",
         "Private Access"
       ],
+      "title": "AccessTypeEnum",
```
anngvu, line 14: So one small annoyance with the LinkML JSON schema generator compared with the previous `make` steps is that `titles` aren't as nice.

sync-to-core-models.py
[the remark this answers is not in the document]
anngvu: Suggest putting it in another branch for development since it seems like a test version still and could maybe be part of #637.

PR #673, opened 2025-07-25
Convert workflow update

anngvu:
Streamlining and fixes for new workflow script. Also other things we need to do that makes sense, such as:
- Moving `super_rules.json`
- Remove current `Makefile` conversion steps that will be handled under new workflow (except for a special one)
- Prepare step for change in `.github/workflows` (note: only removed outdated commands, you're still going to have to add the new dependency installs and new commands for #637)


anngvu:
@cconrad8 Going to merge this so we can facilitate changes in https://github.com/nf-osi/nf-metadata-dictionary/tree/batch-convert more quickly

Issue #634, opened 2025-06-23
Research spike -- test conditional file schema based on assay value


anngvu:
Depends on JSON schemas first generated for file templates (through DPE's approach in new-schematic) or alternative approach #622. Once one of these is available, we can add a rule to materialize the appropriate schema for child files.

PR #674, opened 2025-07-28
Add doc for general dev context and JSON-specific dev

anngvu:
Add some helpful docs about JSON dev and tested behavior on Synapse. Related to #634. Will benefit those the most who've been *less involved* with the JSON dev so far.


anngvu:
So if anyone has anything else that they would like me to cover, create an issue, but I'll go ahead and merge this. Also feel free to add relevant tips and findings to this new dev doc.

PR #689, opened 2025-08-11
Add mappings

anngvu:
@ajs3nj FYI, this is an updated mapping file for CTF. It controls the selection of clinical data elements that we will export to cBioPortal. For the already public CTF tables, pretty sure we include everything. However, for the other CTF clinical table that is under AR, we may need to see if more selection/filtering is desired.


ajs3nj (approved):
Looks good to me!

PR #690, opened 2025-08-13
Update Portal.yaml

cconrad8:
I noticed Data Series and Data Catalog are being populated with NF-OSI Processed Data and the data catalog URL, in cases where it is not applicable.  I'm wondering if this is because there is only a single permissible value?  Added another value (Not Applicable) . 

Also fixed some minor formatting issues.


anngvu:
Fixed some small merge conflicts. Thanks! -- let's how this works.

Issue #692, opened 2025-08-18
Add property to PortalStudy schema to link study phases


anngvu:
This relates to the current NTAP request to link projects, where the second one is an extension. In the past, we already decided to create new projects for the same study (same PIs, research topic) with different funding/contract phases (different timeline and deliverables wording). There's a conversation somewhere about it making it easier to release data this way. For example, see:
- [GFF projects with a Phase B](https://github.com/nf-osi/dcc-site/pull/224)
- DHART projects

So this is distinct from `relatedStudies` -- not sure what it should be called, exactly, but @changtotheintothemoon and @cconrad8 can decide. Then we should fill in this info and update the portal UI so the linkage is more explicit.

anngvu:
Forgot I had a version of this and probably have a better idea of past projects should be linked, so reassigning to me.

Issue #525, opened 2024-09-25
Add `dataCoordinatingCenter`* or similar annotation to Portal Study schema


anngvu:
Use cases:

- Filtering and selection for data cap and cost management as presented by Mackenzie in [this meeting](https://docs.google.com/document/d/1nZGLRKW5LXpY-LBrtrgs4MyO-fb0kDDeouEOvW36xo0/edit#heading=h.mzgscejhs9ke)
- Requests to add ARs can potentially use the metadata to provide more context in the governance workflow. **See https://sagebionetworks.jira.com/browse/PLFM-8553**.

*I don't think we've quite agreed what the attribute would be called.

@cconrad8 for additional thoughts.

cconrad8:
I think "isPartOf" could be added - because isPartOf is more generic. 
Hypothetically there might be studies non-dcc related that we want aggregated, or test projects. 

What do you think?

anngvu:
Confirm with Kevin

anngvu:
Actually, saw highly relevant post [here](https://sagebionetworks.slack.com/archives/C0AFCTGEQ/p1738955317323919) that suggests that platform already made the decision in 2018 (https://github.com/Sage-Bionetworks/SynapseWebClient/commit/2f267c2f946767afd95d86b398b570457b2eafa1) that "Consortium" would be the preferred key and used for indexing, etc. So "Consortium" is basically "Data Coordinating Center". But DCC became more of the standard terminology at Sage in 2022, with DCC playbook, etc. We do actually have consortium in the data model.

anngvu:
I think I closed this because we kept thinking "let's wait on alignment, etc.", but reopening this to work in conjunction with #692

PR #697, opened 2025-09-03
Update NF study meta definition

anngvu:
Resolve some outstanding ideas in #692 and #525.

Note: skipping tests because this is not a manifest template.

### DIFF
--- a/dist/NF.yaml
+++ b/dist/NF.yaml
@@ -7827,13 +7827,13 @@ classes:
         required: false
         description: Synapse project ID of a follow-up project that represents a subsequent funding phase, extension, or continuation of this study's research effort.
         exact_mappings:
-          - successorProject
+          - https://schema.org/successorOf
       previousPhaseProject:
         range: string
         required: false
         description: Synapse project ID of the previous project that this study continues or extends, typically representing an earlier funding phase of the same research effort.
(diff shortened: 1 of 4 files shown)


dist/NF.yaml
```diff
@@ -7817,6 +7822,18 @@ classes:
         multivalued: true
         range: string
         required: false
+      nextPhaseProject:
+        range: string
+        required: false
+        description: Synapse project ID of a follow-up project that represents a subsequent funding phase, extension, or continuation of this study's research effort.
+        exact_mappings:
+          - successorProject
```
allaway, line 7830: https://schema.org/successorOf

dist/NF.yaml
```diff
@@ -7817,6 +7822,18 @@ classes:
...
+      previousPhaseProject:
+        range: string
+        required: false
+        description: Synapse project ID of the previous project that this study continues or extends, typically representing an earlier funding phase of the same research effort.
+        exact_mappings:
+          - predecessorProject
```
allaway, line 7836: https://schema.org/predecessorOf

allaway (approved):
These changes look good to me, just made some suggestions for schema.org mappings that you could use instead/in addition

Issue #756, opened 2025-11-14
Refactor enums to support more contextual templates


anngvu:
Previously, contextually defining different ranges for a property would have required adding custom indirection code on our part because it was not natively supported. But given the current evolution, things will be simpler now and we should put more effort into better set definitions for some of the enums, to help transition to more contextual templates. Starting enum TODOs:

- [ ] AssayEnum: Suggest we minimally break this down into sequencing assays, imaging assays, etc. That will allow us to plug in the correct set as range for the property `assay` for the relevant assay-based templates and make drop-down selection easier on human/AI contributors.
- [ ] PlatformEnum: Similar to above.
- [ ] FileFormatEnum: Similar to above.

Issue #470, opened 2024-06-20
Add data model analysis


anngvu:
Add analysis of data model changes in each PR that:
1. Identifies the number of/which manifest templates affected by this change.
2. Summarizes when keys may collide across files when schematic converts everything to uppercase (i.e. in the final output there is no distinguishing keys that were originally `meningioma` vs `Meningioma` in the source), which can lead to bugs. Also, when editing YAML in most code editors, using a duplicate key under the same parent is clearly pointed out, but duplicate keys can still happen across files. See #450.  
3. Summarizes changes in numbers of props, classes, and enums.

The first is relevant for #312.

PR #761, opened 2025-12-01
Contextual range refactor

anngvu:
Add enum refactor for #756, data model analysis workflow for #470.

Update:
Adding nearly everyone for review since significant structural and syntax changes to be aware of.


BelindaBGarana (approved):
Looks good but didn't test it myself

PR #778, opened 2025-12-22
Migrate conditional requirements

### LINKED ISSUE #691: bug with gene perturbation type?
https://github.com/nf-osi/nf-metadata-dictionary/blob/44a4ebc4c2a09fbabddc8d71137788b623072ae4/modules/props.yaml#L436C1-L436C42

I think this should be knockdown, knockout, etc. At least, that's what it used to be.

### LINKED ISSUE #755: Migrate to new "conditionally required" syntax
This is an update in the same vein as #753, though less critical. In schematic, conditionally required properties uses `requiresDependency`.

**Listing**
```
age:
    annotations:
      requiresDependency: ageUnit
      validationRules: num
    description: A numeric value representing age of the individual. Use with `ageUnit`.
    range: float
    required: false
```

In LinkML, this is called ["skip logic"](https://linkml.io/linkml/howtos/skip-logic.html). Note that this is somewhat different from #426, which conceptually is more like subclass overriding. The Synapse grid has recently finalized feature-parity support for this, thus now is a good time to migrate current instances as well as analyze where else to add new conditional requirements.

BelindaBGarana:
Migrates existing requirements from schematic 'requiredDependency' into new LinkML 'skip logic' to fix #755, maintaining backwards compatibility
Also:
1. ✅ Fixed genePerturbationType enum (was using wrong enum; fixes #691)
2. ✅ Added missing GenomicReferenceEnum to genomicReference slot
3. ✅ Expanded GenomicReferenceEnum with mouse (mm10, mm39) and rat (rn6) references
Completed with AI assistance from Claude Code

### DIFF
--- a/docs/linkml-rules-migration.md
+++ b/docs/linkml-rules-migration.md
@@ -172,6 +172,23 @@ Rules defined in abstract base classes (like `BiologicalAssayDataTemplate`) are 
 4. **Contextual:** Rules defined where slots are used, making relationships clear
 5. **Self-documenting:** Rule descriptions explain the dependency logic
 
+## Enum Validation for Conditional Dependencies
+
+All slots involved in conditional dependencies now have proper enum validation to ensure data quality:
+
+| Slot | Enum | Values |
+|------|------|--------|
+| genePerturbationType | GenePerturbationEnum | knockdown, knockout, overexpression, non-targeting control |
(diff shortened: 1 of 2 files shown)


anngvu (approved):
Nice, especially as `value_presence` turned out to be less obvious, requiring hunting around in the LinkML docs a bit for implementation.

PR #790, opened 2026-01-12
Relax schema constraints and add DCA compatibility

### LINKED ISSUE #783: Relax check constraints for string enums
Per https://sagebionetworks.jira.com/browse/SWC-7657?atlOrigin=eyJpIjoiZTJmNWQzZjEzNzFkNDEwNGJmMTAwYzU0NDRmMmNmMGEiLCJwIjoiaiJ9, need to relax maxLength and maxItems here: https://github.com/nf-osi/nf-metadata-dictionary/blob/85514a147ea4d8a481b9d6caf2c09917764a4837/utils/json_schema_entity_view.py#L154

@anngvu pointed out 'one thing to consider, though, is that we still can't set it too high because the the fileview has total column size limits. I have a script somewhere that does this calculation...'

### LINKED ISSUE #788: Curation task creation incompatible with DCA
If you create a curation task for a folder (bind a specific assay template), that folder is no longer visible/annotatable in the Data Curator App (which requires superdataset template)

BelindaBGarana:
## Summary
- Fixes #783: Relax check constraints for string enums in fileview creation
- Fixes #788: Add contentType='dataset' to all templates for DCA compatibility

## Changes
**Schema constraints** (utils/json_schema_entity_view.py):
- Increased STRING column max_size: 100 → 250 characters
- Increased STRING_LIST column max_size: 5 → 100 characters  
- Increased enum value limit: 100 → 1000 values
- Added support for nested enums in list-type columns

**DCA compatibility** (all 48 template schemas):
- Added contentType='dataset' conditional to keep folders visible in Data Curator App after binding assay templates

## Notes
The new limits accommodate the modelSystemName field in ImagingAssayTemplate (809 enum values, up to 56 chars each). **We are not certain about Synapse fileview total column size limits** - these values are conservative estimates based on the 64KB total limit, but may need adjustment if issues arise.

### DIFF
--- a/registered-json-schemas/AnimalIndividualTemplate.json
+++ b/registered-json-schemas/AnimalIndividualTemplate.json
@@ -1024,13 +1024,6 @@
           "concreteType": {
             "const": "org.sagebionetworks.repo.model.Folder"
           }
-        },
-        "not": {
-          "properties": {
-            "name": {
-              "const": "Raw Data"
-            }
-          }
         }
       },
       "then": {
--- a/registered-json-schemas/BehavioralAssayTemplate.json
+++ b/registered-json-schemas/BehavioralAssayTemplate.json
@@ -116,13 +116,6 @@
           "concreteType": {
             "const": "org.sagebionetworks.repo.model.Folder"
           }
-        },
-        "not": {
-          "properties": {
-            "name": {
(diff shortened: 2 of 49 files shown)


anngvu (changes requested):
Small change suggested

PR #803, opened 2026-01-22
Fix Synapse schema limits and constraints

BelindaBGarana:
## Summary
- Fix 64KB row size limit by reducing column sizes and adding maximum_list_length constraints
- Remove enum_values from entity view columns to stay within row size limits
- Implement conditional enum filtering for modelSystemName field
- Add automatic deletion of existing file views before recreation
- Fix handling of nullable types in JSON Schema entity view creation

## Test plan
- [x] Verify entity views create successfully without 64KB errors
- [ ] Confirm enum filtering works correctly for model systems - might need to wait until schemas are re-registered with Synapse upon merge to main
- [x] Test file view recreation with auto-deletion


anngvu (commented):
Yeah, I think this is too complicated for now, unfortunately. I'm just going to bring over the essence of what's really needed to make things work to PR #816, and close this.

PR #833, opened 2026-03-09
Add spatial transcriptomics template (#160)

### LINKED ISSUE #160: Add template+standards for spatial transcriptomics data
We have at least one dataset coming in that uses ST. Should assess what metadata to collect and what file formats to require. There's both imaging and transcriptomics components, so this might require multiple templates.

HTAN likely has a draft spec, but not sure.

BelindaBGarana:
## Summary
- Adds `SpatialTranscriptomicsTemplate` in `modules/Template/Data_SpatialTranscriptomics.yaml`, modeled after the `sagedm-spatial` Synapse schema
- Adds three new slots to `props.yaml`: `cytassistSample`, `probeSet`, `visiumslideID`
- Adds `10X Visium CytAssist` to `SequencingPlatformEnum` in `modules/Assay/Platform.yaml`

## Template details
- Inherits from `GeneticsAssayTemplate` (provides specimen, diagnosis, tumorType, specimenPreparationMethod, etc.)
- Assay defaults to `spatial transcriptomics`; dataType defaults to `gene expression`
- Platform accepts any of `SequencingPlatformEnum`, `ArrayPlatformEnum`, or `ImagingSystemPlatformEnum` to cover 10x Visium, CytAssist, NanoString CosMx/GeoMx
- FileFormat accepts sequencing (fastq, bam, etc.) or imaging (tiff, ome-tiff) formats

## Test plan
- [ ] Verify `SpatialTranscriptomicsTemplate` is generated correctly in NF.jsonld
- [ ] Confirm new slots (`cytassistSample`, `probeSet`, `visiumslideID`) appear in the schema
- [ ] Confirm `10X Visium CytAssist` appears as a valid platform value

Closes #160

### DIFF
--- a/modules/Template/Data_SpatialTranscriptomics.yaml
+++ b/modules/Template/Data_SpatialTranscriptomics.yaml
@@ -1,0 +1,41 @@
+classes:
+  SpatialTranscriptomicsTemplate:
+    annotations:
+      required: false
+      requiresComponent: ''
+      templateUsage: most_common
+      templateFor:
+        dataType:
+        - gene expression
+        assay:
+        - spatial transcriptomics
+      dataGranularity: specimen-level
+    close_mappings:
+    - htan:VisiumSpatialTranscriptomicsLevel1
+    description: Template for describing raw data from spatial transcriptomics experiments,
+      including sequencing-based approaches (e.g. 10x Visium, 10x Visium CytAssist)
+      and in-situ hybridization-based approaches (e.g. NanoString CosMx, GeoMx).
+    is_a: GeneticsAssayTemplate
+    slots:
+    - captureArea
+    - visiumslideID
+    - cytassistSample
+    - probeSet
+    - libraryPrep
+    - libraryPreparationMethod
+    - specimenType
+    slot_usage:
+      assay:
+        range: SequencingAssayEnum
+        ifabsent: string(spatial transcriptomics)
+      platform:
+        any_of:
+        - range: SequencingPlatformEnum
+        - range: ArrayPlatformEnum
+        - range: ImagingSystemPlatformEnum
+      fileFormat:
+        any_of:
+        - range: SequencingFileFormatEnum
+        - range: ImagingFileFormatEnum
+      dataType:
+        ifabsent: string(gene expression)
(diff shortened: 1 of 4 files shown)


allaway:
Sorry for my delay @BelindaBGarana - i have a fairly basic first question - should we consider having different schemas for the imaging vs omics portion of the spatial data? would that make it easier to enforce minimal metadata for both (rather than having a merged schema). it also seems like other repositories do modality-specific templates, so I guess that's another option to consider. what do you think ?

BelindaBGarana:
> Sorry for my delay @BelindaBGarana - i have a fairly basic first question - should we consider having different schemas for the imaging vs omics portion of the spatial data? would that make it easier to enforce minimal metadata for both (rather than having a merged schema). it also seems like other repositories do modality-specific templates, so I guess that's another option to consider. what do you think ?

I like that idea @allaway - just confirming that I split up the imaging and sequencing schemas

modules/Template/Data_SpatialTranscriptomics.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - libraryPreparationMethod
+    - specimenType
+    slot_usage:
+      assay:
+        range: SequencingAssayEnum
+        ifabsent: string(spatial transcriptomics)
+      platform:
+        any_of:
+        - range: SequencingPlatformEnum
+        - range: ArrayPlatformEnum
```
allaway:
```suggestion
```

modules/Template/Data_SpatialTranscriptomics.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - range: ImagingSystemPlatformEnum
```
allaway:
```suggestion
```

allaway: Maybe you had a reason to keep these that I am not thinking of!

allaway (changes requested):

allaway:
Couple of minor comments, but looks good to me otherwise! once those changes are made no need for a re-review

PR #834, opened 2026-03-09
Fix enum-template validation mismatches for Synapse Curator (#831)

### LINKED ISSUE #831: Do validation analysis after majority Synapse Curator rollout
**Context**

With the transition from DCA to Synapse Curator, we need to do an analysis for potential issues with existing data because the two systems can work with the data model differently. Below are several examples already discovered in an analysis after a small early setup batch (~10% rollout). But an analysis after majority rollout (>50% projects) should uncover more issues. 

- For some existing annotations, "age" might not validate in Synapse even though the data is there and looks fine initially. But it turns out Synapse won't validate these because they were submitted as string at some point by schematic. For other int/num attributes, we had to temporarily type as string as a workaround to unblock submission (see https://github.com/nf-osi/nf-metadata-dictionary/issues/151).  
- Synapse validation can include the `manifest.csv` as well, while with schematic it does not (file is uploaded after validation). In general, we need better data model rules to dynamically tailor validation, somewhat captured in #825.
- Synapse Curator includes a new feature where enums are more constrained/contextually relevant for their template, achieved through enum refactoring. Discovered issue where "western blot" wasn't in the assay enum set for ImagingAssayTemplate, which unnecessarily failed existing data. There are likely similar instances.

**Goal**

This analysis will inform data model work still needed -- or sometimes data migration work (such as string -> num). Results from this are also important for [#134](https://github.com/nf-osi/jobs/issues/134), to determine reasonable initial monitoring parameters. Some validation issues are our issues, not the contributors'.

BelindaBGarana:
## Summary
Fixes validation failures discovered during the Synapse Curator majority rollout where `templateFor.assay` values were missing from the enum range assigned in each template's `slot_usage.assay`. When Synapse validates data against a template, it rejects assay values not in the template's enum — causing failures for existing valid data.

**Enum additions (`modules/Assay/Assay.yaml`)**
- `ImagingAssayEnum`: added `gel filtration chromatography` and `high content screen` (both were only in `CellBasedAssayEnum` but referenced by `ImagingAssayTemplate`)
- `SequencingAssayEnum`: added `next-generation sequencing` (referenced by `BulkSequencingAssayTemplate`, `ScSequencingAssayTemplate`, and others but absent from the enum)
- `MassSpectrometryAssayEnum`: added `SomaScan` (referenced by `AffinityProteomicsTemplate` but absent from the enum)
- `ClinicalBehavioralAssayEnum`: added `survival` (referenced by `ClinicalAssayTemplate` but was only in `CellBasedAssayEnum`)
- `CellBasedAssayEnum`: added `pattern electroretinogram` (referenced by `ElectrophysiologyAssayTemplate` but was only in `ClinicalBehavioralAssayEnum`)

**Template fixes (`modules/Template/Data_CellAssays.yaml`)**
- `CellTissuePhenotypingTemplate`: replaced alias `Matrigel-based tumorigenesis assay` in `templateFor.assay` with the canonical enum value `in vitro tumorigenesis`
- `PlateBasedReporterAssayTemplate`: removed `2D Incucyte` from `templateFor.assay` — it is a platform (in `OtherPlatformEnum`), not an assay value

## Test plan
- [ ] Verify NF.jsonld builds cleanly with no enum errors
- [ ] Confirm the 5 new assay enum values are visible in Synapse Curator dropdowns for the relevant templates
- [ ] Spot-check existing NF Portal data with `assay=high content screen`, `assay=survival`, `assay=pattern electroretinogram` etc. now passes validation

Closes #831

### DIFF
--- a/modules/Assay/Assay.yaml
+++ b/modules/Assay/Assay.yaml
@@ -66,6 +66,12 @@ enums:
       miRNA-seq:
         description: Small RNA measurements collected from RNA sequencing experiments
         meaning: OBI:0002112
+      next-generation sequencing:
+        aliases:
+        - NGS
+        - next generation sequencing
+        description: High-throughput DNA sequencing methods that allow massively parallel sequencing of millions of DNA fragments simultaneously.
+        meaning: NCIT:C101294
       next generation targeted sequencing:
         aliases:
         - Targeted NGS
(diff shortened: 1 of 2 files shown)


anngvu (approved):
Thanks!!

PR #864, opened 2026-04-15
remove ordering

cconrad8:


anngvu (commented):
We reviewed and this doesn't change much for now; focus will be on upstream fix or implementing something that replaces `gen-json-schema` usage entirely

PR #875, opened 2026-04-23
Add --replace mode and GenericDataResourceTemplate to curation task workflow

BelindaBGarana:
## Summary

- Adds `GenericDataResourceTemplate` to the workflow template dropdown for cases where file format (e.g. Excel) doesn't fit a pipeline-output-specific template
- Adds a `replace` boolean input to the workflow and `--replace` flag to `create_curation_task.py` to safely change the template on an already-configured folder — handles unbinding the old schema and deleting the existing curation task before creating a new one
- Before any destructive action, warns if files in the folder already have annotations matching the new template's fields (checks up to 10 files; ignores system fields like `createdBy`)

## Test plan

- [x] Successfully tested replacing WES template with `GenericDataResourceTemplate` for folder `syn74327086`: [run #24852688524](https://github.com/nf-osi/nf-metadata-dictionary/actions/runs/24852688524)
- [ ] Test `--replace` / `replace=true` on a folder with an existing curation task to verify old task is deleted and schema is rebound
- [ ] Verify pre-annotation warning fires correctly when files already have matching fields

### DIFF
--- a/.github/workflows/create-curation-task.yml
+++ b/.github/workflows/create-curation-task.yml
@@ -15,6 +15,9 @@ on:
           # Default option - most commonly used for genomics data
           - GenomicsAssayTemplate
 
+          # Generic fallback (use when file format doesn't fit a specific pipeline-output template)
+          - GenericDataResourceTemplate
+
           # Assay technical metadata templates
           - BehavioralAssayTemplate
           - BiologicalAssayDataTemplate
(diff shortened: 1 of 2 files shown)


anngvu:
Tested with https://github.com/nf-osi/nf-metadata-dictionary/actions/runs/25005644309

anngvu (approved):
LGTM!

PR #877, opened 2026-04-27
Add shRNA-seq to SequencingAssayEnum

allaway:
## Summary

- Adds `shRNA-seq` (EFO:0030034) to `SequencingAssayEnum` in `modules/Assay/Assay.yaml`
- Includes alias `shRNA library screen sequencing` and a description of the method
- Term was needed to annotate the dataset in nf-osi/nadia#35 (ArrayExpress E-MTAB-8398, Synapse syn74280931), an shRNA library screen using amplicon sequencing

## Test plan

- [ ] Confirm `shRNA-seq` appears correctly in built artifacts after CI runs
- [ ] Verify EFO:0030034 resolves to the correct ontology term


anngvu (approved):
LGTM -- glad that meaning URI is actually valid

allaway:
@anngvu  only because I told it what the uri was ;) https://github.com/nf-osi/nadia/issues/35#issuecomment-4306690945

claude _can_ get it right if pushed, I have gotten it to actually search for the uri when it got it wrong the first time, but ideally it would just do that in the first pass

PR #887, opened 2026-05-07
Add release agent

anngvu:
This implements transition to agentic release (brought up [in this sprint review](https://docs.google.com/document/d/1Svq4QwwjmG6BmAlR5hG8_rtM7sIv_HUTod2hpf9igww/edit?tab=t.0#heading=h.fo2tnzvdf9yn)). https://github.com/nf-osi/nf-metadata-dictionary/releases/tag/v10.7.13 is our first agentic release. There's still an option to manually create a release version outside of the automated agent schedule.


.github/workflows/release-new-version.yaml
```diff
@@ -1,120 +1,259 @@
...
+            what has been officially released, not the schema versions alone.
+
+          Respond with ONLY valid JSON, no markdown fences or extra text:
+          {"release": true, "version": "MAJOR.MINOR", "reasoning": "one sentence", "notes": "markdown release notes"}
+          or
+          {"release": false, "reasoning": "one sentence"}
+          PROMPT
+
+          PAYLOAD=$(jq -n --rawfile prompt /tmp/prompt.txt '{
+            model: "claude-haiku-4-5",
```
changtotheintothemoon: should we use sonnet 4.6?
haiku 4.5 sounds weak IMO...

allaway:
LGTM :shipit: 



I think it would be fine to change it to sonnet as James mentioned but if it works fine w/haiku,  that's gerat too

.github/workflows/release-new-version.yaml
anngvu: Upgraded per suggestion

Issue #840, opened 2026-03-25
Remove schematic test suite and other no-longer-relevant schematic workflows


anngvu:
In the post Synapse Curator MVP-era, we can start dismantling the schematic workflows, starting with the schematic test suite, which is one of the more burdensome tests.

PR #893, opened 2026-05-19
Migrate from schematic

anngvu:
We are shutting down DCA next week and the schematic API service might soon be taken down as well (keeping schematic API + DCA up does cost more than one would expect). This means we actually can't run some parts of the schematic CI pipeline, and it doesn't make sense anyway to support a second set of workflows for something that's not used. Thus there are two related items this addresses, detailed below.

#### Issue #840 — Remove no-longer-relevant schematic workflows for test, build, etc.                            
                                                                                                                                                                                                       
  - ✅ All of `tests/schematic/` removed
  - ✅ Schematic CI steps removed from all three workflows (main-ci, rebuild-artifacts-on-main,                   
  release-new-version)                                                                                          
  - ✅ Schematic-related utils removed + update or confirm no downstream callers (`schematic-flavored-linkml-*.json`, `fix_display_names.py`, `utils/enumerate.sh`)                           
  - ✅ `retold` removed from build and CI contexts since it was another helper to generate JSON-LD (Makefile, devcontainer, all three workflows)     
  - ✅ `dispatch-downstream-staging.yml` was already disabled, now also deleted (dispatched to old DCA repo)                                                          
  - ✅ README updated to reflect legacy status of schematic and NF.jsonld                                        
  - ✅ NF.jsonld removed from artifact build targets and auto-commit                                              
                                                                                                                  
Remaining for now: `requiresDependency` annotations in `props.yaml` for reference. 
                                                                                                                                                                                                                             
####  Issue #890 — "Full-parity or better" test suite re-implementation for LinkML                                                                               

  - ✅ 32+ JSON fixtures across 8 templates replacing the old schematic CSV fixtures
  - ✅ 4 test registries: age masking, genomics, required fields, conditional validation                          
  - ✅ `test_template_datatypes.py` verifies dataType annotation coverage across all templates
  - ✅ Pytest wired into CI with results posted back to PR as a comment                                           
  - ✅ Overall, exceeds old schematic test breadth


changtotheintothemoon (approved):
lgtm! will visit often to see changes toward CI

Issue #825, opened 2026-02-26
Apply relaxed conditional validation for co-located auxiliary files like `manifest.csv` `.out` `.log`


anngvu:
It's common to have files like `manifest.csv`, `.out`, `.log`, `.sh` _within_ the dataset folder with the primary data (see example below), but the validation for these files should be different (make fewer "required"). This helps to:
- Reduce burden of having to add annotations that don't make semantic sense or having to move the files
- Make validation stats more accurate / reduce erroneously sending out notifications that there are still validation errors for some files 

See if conditional rules afforded by JSON schema can be used to , e.g. based on `resourceType`, `dataType`, `fileFormat` or something else.

<img width="400" height="380" alt="Image" src="https://github.com/user-attachments/assets/e01dc507-22d2-43b1-9750-a85ad2725994" />

anngvu:
Approach is good. You won't need below since LinkML will generate the conditionals in the schema without us needing to update our script.

> Update validation rules in utils/gen-json-schema-class.py

PR #901, opened 2026-05-29
Term and template revisions, additions for #889

anngvu:
See #889 - a bit of a hodgepodge: 

- Many term additions and making sure templates have access to the file enums they need. 
- For @changtotheintothemoon, relevant to your issue #825, this does add `manifest` to `resourceType` since it's special enough to label more specifically than `metadata` - so you can relax validation for `manifest` as well as the other ones already planned. 
- Relaxed `platform` req for PlateBasedReporterAssayTemplate (where it's not as crucial as for sequencing/imaging), but on the other hand became a little stricter on requiring `expressionUnit` for processed data. 
- Added two templates to better handle other processed/result data.


changtotheintothemoon (approved):
LGTM

PR #903, opened 2026-06-03
Add Synapse search integration docs

anngvu:
Went through and refined docs overall, but the main focus is really adding notes and test examples for the new Synapse search management services. Review should mostly be on `dev/SYNAPSE_SEARCH.md` to confirm things are understandable in terms of the integration picture and what we'll need to do next. Another important connection to make: @changtotheintothemoon's earlier effort in importing synonyms will now realize greater positive impact :grin:, since that result will actually be applied in the new search system.


changtotheintothemoon (approved):
lgtm will monitor!

PR #908, opened 2026-06-04
Fix and reconfigure synonyms workflow

changtotheintothemoon:
## Summary

Fixes the synonyms workflow that has been failing due to timeouts (#905).

- **Remove `release` trigger** — workflow now runs on-demand only via `workflow_dispatch`
- **Replace broken ontology fallback with OLS4 API** — the old `fetch_full_ontology()` tried to download entire `.owl` files (e.g. NCIT.owl ~500MB) with a 15s timeout, which always failed. Now uses the OLS4 REST API as primary method (~0.5s per term), with RDF content negotiation as fallback
- **Reduce timeouts** — workflow: 70min → 20min, shell: 60min → 15min, script: 55min → 10min
- **Fix injection to merge aliases** — previously replaced all existing aliases with only ontology-derived ones, destroying manually-added backward-compatibility aliases. Now merges new synonyms with existing ones
- **Add `--cleanup` flag** to `inject_synonyms.py` for removing case/spacing-only duplicate aliases
- **Remove 69 low-quality aliases** across 4 module files where the alias differed from the term only by case, spacing, or punctuation (e.g. "next generation sequencing" as alias of "next-generation sequencing")

### Performance improvement

| Metric | Before | After |
|--------|--------|-------|
| Extraction time | 45+ min (timeout) | ~3 min |
| Workflow timeout | 70 min | 20 min |
| API method | RDF content negotiation + .owl download | OLS4 REST API (JSON) |

## Test plan

- [x] Validated all modified YAML files parse correctly
- [x] Tested OLS4 extraction locally against 10 terms — synonyms returned in 3.7s
- [x] Estimated full extraction: ~150s for 399 ontology-mapped terms
- [x] Verified `inject_synonyms.py --cleanup --dry-run` identifies correct aliases
- [ ] Run `workflow_dispatch` with `action=test-inject` after merge to verify end-to-end

Closes #905

### DIFF
--- a/.github/workflows/synonyms.yml
+++ b/.github/workflows/synonyms.yml
@@ -1,10 +1,7 @@
 name: Synonyms Management
 
 on:
-  release:
-    types: [published]
-  
-  workflow_dispatch:  # Allows manual triggering
+  workflow_dispatch:
     inputs:
       action:
         description: 'Action to perform'
@@ -30,15 +27,13 @@ on:
 jobs:
   synonyms:
     runs-on: ubuntu-latest
-    timeout-minutes: 70  # Allow enough time for extraction
+    timeout-minutes: 20
     permissions:
       contents: write  # Allow the workflow to push changes back to the repository
       pull-requests: write  # Allow creating pull requests
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
-          ref: main  # Always checkout main branch, even when triggered by release
 
       - name: Print workflow info
         run: |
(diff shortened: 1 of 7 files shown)


anngvu (changes requested):
Nice, huge improvements so far! One change: In the test run that created PR#909, noticed that we have unnecessary rewritten files for `AnimalModel.yaml` and `CellLineModel.yaml` even though there are no synonym changes in these files. There is an unconditional write in `inject_synonyms_into_yaml()` so a file is rewritten even when `modifications_made == 0`. It'll be a bit cleaner/easier to review if the write checked something like `if modifications_made > 0`.

changtotheintothemoon:
@anngvu Addressed both issues in the latest push (c00705f):

**1. Unnecessary file rewrites**
`inject_synonyms_into_yaml()` now checks `if modifications_made > 0` before writing, so files like AnimalModel.yaml and CellLineModel.yaml won't be touched when there are no synonym changes.

**2. Wrong meaning IRI for NGS (#910)**
Fixed `next-generation sequencing` mapping from `NCIT:C101294` (Whole Genome Sequencing) → `NCIT:C101293` (Next Generation Sequencing). Verified via OLS4:
- C101293 label: "Next Generation Sequencing", synonyms include "NGS", "Massively Parallel Sequencing"
- C101294 label: "Whole Genome Sequencing", synonyms include "WGS"

anngvu (approved):
Thanks, LGTM now!

PR #874, opened 2026-04-20
Constrain model* attributes to PdxGenomicsAssayTemplate

BelindaBGarana:
Closes #872

## Summary
- Remove `modelSystemName`, `modelSpecies`, `modelSex`, `modelAge`, `modelAgeUnit` from `BiologicalAssayDataTemplate` so they no longer appear in all data templates
- Add `modelSpecies`, `modelSex`, `modelAge`, `modelAgeUnit` explicitly to `PdxGenomicsAssayTemplate` (which already had `modelSystemName`)
- Templates that explicitly declare `modelSystemName` (`BehavioralAssayTemplate`, `CellTissuePhenotypingTemplate`) are unaffected

## Test plan
- [x] Verify `PdxGenomicsAssayTemplate` JSON schema includes all `model*` fields
- [x] Verify other data templates (e.g. `RNASeqTemplate`) no longer include `model*` fields
- [x] Verify `BehavioralAssayTemplate` and `CellTissuePhenotypingTemplate` still include `modelSystemName`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #934, opened 2026-07-01
Validate migration of and remove schematic annotations for deps, validation rules

anngvu:
Close #929. An earlier migration deferred removing schematic notations in the data model, which this now removes and finalizes transition with schematic backward-compatibility. Before removing the annotations, validate that valuable schematic dep/validation rules have been re-implemented in LinkML. Found and fixed only one place where `genePerturbed` dep rule was not implemented. Add tests.

Update: Add a fix for #935

### DIFF
--- a/modules/Template/Biosample.yaml
+++ b/modules/Template/Biosample.yaml
@@ -5,10 +5,8 @@ classes:
       requiresComponent: ''
       templateUsage: most_common
       dataGranularity: individual-level
-    description: 'Template for non-human individual-level data. For xenografts, individualID
-      references the cell line donor. Use sex/age/species for the individual animal.
-      Use modelSystemName for the animal model genetic background, with modelSpecies/modelSex/modelAge
-      for the model host animal.'
+    description: 'Template for non-human individual-level data. Use sex/age/species
+      for the individual animal, and modelSystemName for the name of the mouse model.'
     slots:
     - individualID
     - sex
@@ -23,10 +21,6 @@ classes:
     - germlineMutation
     - species
     - modelSystemName
-    - modelSpecies
-    - modelSex
-    - modelAge
-    - modelAgeUnit
   BiospecimenTemplate:
     annotations:
       required: false
--- a/modules/Template/Biosample.yaml
+++ b/modules/Template/Biosample.yaml
@@ -46,6 +46,16 @@ classes:
     - modelSex
     - modelAge
     - modelAgeUnit
+    rules:
+    - description: If modelAge is provided, modelAgeUnit must be provided
+      preconditions:
+        slot_conditions:
+          modelAge:
+            value_presence: PRESENT
+      postconditions:
+        slot_conditions:
+          modelAgeUnit:
+            value_presence: PRESENT
   HumanCohortTemplate:
     annotations:
       required: false
--- a/modules/Template/Biosample.yaml
+++ b/modules/Template/Biosample.yaml
@@ -27,6 +27,16 @@ classes:
     - modelSex
     - modelAge
     - modelAgeUnit
+    rules:
+    - description: If modelAge is provided, modelAgeUnit must be provided
+      preconditions:
+        slot_conditions:
+          modelAge:
+            value_presence: PRESENT
+      postconditions:
+        slot_conditions:
+          modelAgeUnit:
+            value_presence: PRESENT
   BiospecimenTemplate:
     annotations:
       required: false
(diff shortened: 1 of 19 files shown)


changtotheintothemoon:
## Review

Verified the migration by rebuilding the model locally (`make NF.yaml` → generate `ChIPSeqTemplate` JSON schema → validate the 4 new fixtures). All 4 pass their expected valid/invalid outcomes ✅. Also confirmed the removed annotations are **inert in the current pipeline** — the Makefile strips all `annotations` (`yq 'del(.. | select(has("annotations")))'`) before schema generation — so this removal changes **zero** generated artifacts except the new `genePerturbed` rule. **No functional regression is introduced.**

### 🟠 `modelAge ↔ modelAgeUnit` was never migrated, and this PR removes its last trace

The description says only `genePerturbed` needed a fix, but that isn't complete. Both `modelAge → modelAgeUnit` and `modelAgeUnit → modelAge` have **0 corresponding LinkML rules** across all 9 templates that expose these slots (`AnimalIndividualTemplate`, `BiospecimenTemplate`, `PdxGenomicsAssayTemplate`, `EpigeneticsAssayTemplate`, `PharmacokineticsAssayTemplate`, `ProteinArrayTemplate`, `ProcessedExpressionTemplate`, `GeneralMeasureDataTemplate`, `ImmunoMicroscopyTemplate`). Verified two ways (source grep = 0; artifact `then.required` for `modelAgeUnit` = 0).

This is the direct animal-model analog of `age ↔ ageUnit`, which the model *does* enforce (with an "Unknown" exception), and the field description even states *"Required when modelAge is specified."* So it reads as an oversight rather than a deliberate exclusion.

Nuance: because the annotations were already inert, this dependency was already unenforced before the PR — not a new runtime regression. But since the PR's purpose is to *finalize* the migration after validating coverage, this is a valuable dependency being retired without a replacement.

**Suggest:** either add the `modelAge ↔ modelAgeUnit` rules (mirroring `age ↔ ageUnit`) to the 9 classes here, or explicitly note it's intentionally dropped.

### 🟡 Smaller gaps left unenforced
- `age → ageUnit` is enforced everywhere **except `AnimalIndividualTemplate`** (has both slots, no rule).
- `dataType → dataSubtype` is unenforced in `AnalysisResultTemplate` and `GenericDataResourceTemplate` (both offer the slots, no rule).

Same nuance (pre-existing, annotations were inert) — worth a migrate-or-explicitly-drop decision.

### 🟢 Minor
- `aliquotID`/`parentSpecimenID → specimenID` is now narrowed to `resourceType == experimentalData` for ~21 assay templates rather than "whenever the source is present." Consistent with the #825 auxiliary-file relaxation and likely intentional — noting for completeness.
- Nit: the new `ChIPSeqTemplate` `genePerturbationTechnology` rule description is single-line while the analogous pre-existing rules wrap across lines. Cosmetic.

### What's solid ✅
- **`genePerturbed` migration is complete and correct**: all 6 classes using the slot (`ChIPSeqTemplate`, `RNASeqTemplate`, `ScRNASeqTemplate`, `GenomicsAssayTemplateExtended`, `BehavioralAssayTemplate`, `CellTissuePhenotypingTemplate`) carry both rules; none reference a slot the class lacks.
- **`validationRules` fully covered by native LinkML** (`readPair` = `range: integer, min 1, max 2` = "inRange 1 2"; "list like" → `multivalued: true`; num/int → float/integer ranges).
- **Tests are well-designed** — the `xfail(strict=True)` pattern makes a broken rule surface as XPASS/red, and using `resourceType: metadata` correctly isolates the rule from #825's `experimentalData` requirements. E2E-confirmed passing.

### Verdict
Approve with one requested change: the core `genePerturbed` work is correct, well-tested, and regression-free, but please re-implement (or explicitly declare dropped) **`modelAge ↔ modelAgeUnit`** before finalizing the removal, and decide on the two 🟡 gaps.

anngvu:
> across all 9 templates that expose these slots `(AnimalIndividualTemplate, BiospecimenTemplate, PdxGenomicsAssayTemplate, EpigeneticsAssayTemplate, PharmacokineticsAssayTemplate, ProteinArrayTemplate, ProcessedExpressionTemplate, GeneralMeasureDataTemplate, ImmunoMicroscopyTemplate)`

In the actual model, these two slots are only exposed by 3 templates: AnimalIndividualTemplate, BiospecimenTemplate, PdxGenomicsAssayTemplate. See #874. 

But existence of `modelAge` in AnimalIndividualTemplate is a bug.

PR #938, opened 2026-07-02
Type slots that fall back to default `string` (#930, Phase 1)

changtotheintothemoon:
Addresses #930 (Phase 1). Many top-level slots have no explicit `range` and fall back to `default_range: string`, losing JSON-Schema typing and validation. This is a deliberately **conservative, low-risk first pass**; the remaining slots are handled in follow-ups (see Deferred).

### Recover 9 controlled vocabularies as named enums
These slots documented an inline `enum_range`, but the build **silently strips `enum_range`** (`Makefile`: `yq 'del(.. | select(has("enum_range")).enum_range)'`), so today they emit plain `string` with no enforcement. Promoted to named enums using the **identical values** (no change to accepted values):

`captureArea`, `dataCollectionMode`, `epidemiologyMetric`, `immersion`, `peakCallingAlgorithm`, `programmingLanguage`, `progressReportNumber`, `recordingSource`, `slideVersion`

New enums: 6 assay/experiment ones in `modules/Assay/Parameter.yaml`; `ProgrammingLanguageEnum`, `EpidemiologyMetricEnum`, `ProgressReportNumberEnum` in new small topical files under `modules/Other/`.

### Type numeric metric slots
`float`: `averageBaseQuality`, `averageInsertSize`, `averageReadLength`, `compoundDose`, `concentrationNaCl`, `meanCoverage`, `readsDuplicatedPercent`
`integer`: `targetDepth`
(mirrors already-typed siblings like `readDepth`, `readsMappedPercent`.)

### Type a boolean flag
`isFilteredReads` → `BooleanEnum` (repo house convention models booleans as a Yes/No enum, not native `boolean`).

### Tests
Added `tests/test_registry_slot_types.yaml` + fixtures validating `peakCallingAlgorithm` enum enforcement on `ChIPSeqTemplate` (valid `MACS2` / invalid unlisted value), following the existing `xfail`-strict registry-test pattern.

## Deferred to follow-up PRs (intentionally not here)
- Value-reconciliation enums: `materialType` (vs `Material`), `transplantationRecipientSpecies` (common→scientific names vs `SpeciesEnum`)
- `cytassistSample` (True/False → Yes/No migration), `reporterSubstance` (no vocabulary yet)
- URI typing (`uriorcurie` — no precedent in the model today), Synapse platform-managed system fields (intentionally `string`), dates, and a gene-symbol enum
- A pre-release backward-compatibility audit of existing Synapse annotations before tightening types

## Notes
- Verified locally: merged model builds and every new `range:` resolves (no dangling references); enums/types materialize correctly; no existing `tests/data/**` fixture uses any changed slot (no regression).
- Generated artifacts (`registered-json-schemas/`, `dist/`) are intentionally **not** committed, per repo convention — CI rebuilds and Synapse-validates them, and the `peakCallingAlgorithm` test validates against the CI-regenerated schema.

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -245,6 +245,14 @@ jobs:
           message-id: pytest-report
           message-path: pytest_report.md
 
+      # `pytest | tee` makes the run step exit 0 even when tests fail, so the
+      # captured exit code has to be re-asserted after the report is posted.
+      - name: Fail if pytest failed
+        if: always() && steps.pytest.outputs.exit_code != '0'
+        run: |
+          echo "::error::pytest exited with ${{ steps.pytest.outputs.exit_code }} - see the pytest report comment on this PR."
+          exit 1
+
   check-schema-limits:
     name: Validate schema limits (Synapse platform)
     needs: [build]
--- a/modules/Assay/Parameter.yaml
+++ b/modules/Assay/Parameter.yaml
@@ -428,3 +428,59 @@ enums:
       micrometer:
       millimeter:
       nanometer:
+  CaptureAreaEnum:
+    description: Active region on a Visium/CytAssist slide where a tissue sample can be placed.
+    permissible_values:
+      A:
+      B:
+      C:
+      D:
+      A1:
+      B1:
+      C1:
+      D1:
+  DataCollectionModeEnum:
+    description: Mode of data collection in tandem MS assays.
+    permissible_values:
+      DDA:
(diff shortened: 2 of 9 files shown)


anngvu (changes requested):
Looks good for fixing/enforcing stricter range. Just need to fix the newly added test instance data `ChIPSeqTemplate/valid_peakcalling_macs2.json`

anngvu (approved):
Thanks!

PR #953, opened 2026-08-20
Update curation task utils, building on new synapseclient support

### LINKED ISSUE #952: Review curation task utils (`create_curation_task.py`, `create_recordset_task.py`)
Curation task utils need a review since the last 4 months. Note that these utils were developed when nothing was yet in the Python synapse client; they also include many fixes and NF-specific enhancements that were not in the client (and may still not exist in the client). For example, we had to implement column-size tuning ourselves when creating a fileview to avoid the 64kb row size limit error (see PR #843). 

Review and either build on top of the latest client (counting this a maintenance update) or remove utils and associated workflows entirely if there is no real benefit given the latest infrastructure.

anngvu:
Closes #952. This is an overall improvement building on top of recently released client features; the scripts still include guards/niceties you don't get otherwise. Suggesting @aditya-nath-sage be primary reviewer for this one since relevant to a recent CTF project question, but additional eyes/awareness always welcome. 

#### Changes
- Remove stale `@develop references`. Both `create_curation_task.py` and `create_recordset_task.py` predated synapseclient's released curation-task support and pointed at `pip install git+...@develop`. Replaced with accurate version floors (synapseclient>=4.12.0) in all relevant locations.
- Rebuild `create_curation_task.py` on top of newer client utilities, dropping ~90 lines of now-redundant original code:
  - `synapseclient.extensions.curator.utils.project_id_from_entity_id()` replaces a manual folder→project hierarchy walk.
  - `Folder.bind_schema()` replaces the old JsonSchemaService bind/unbind pair — binding is a PUT, so rebinding a folder with a different schema just overwrites; the manual unbind-before-rebind step was unnecessary and is removed.
  - Manually-constructed id/name Columns (which collided with EntityView's own `include_default_columns` defaults and risked corrupting id's special meaning) are gone in favor of `reorder_column()`.
  - Deliberately _did not_ delegate column creation to the client's own builder — it uses unbounded `MEDIUMTEXT`, while ours caps `STRING` at 80 chars / lists at 20 items to stay under Synapse's 64KB file-view row limit. Using the client's builder as-is would have reintroduced that regression. Note: We need a more optimized version bc our templates have a large number of columns.
- Add new `--assignee` support. Resolves a username/email/team name/numeric ID to a Synapse principal ID and sets it on the CurationTask. Adds a permission-check-only step that warns if the assignee lacks READ/CREATE/UPDATE on the upload folder, which means they wouldn't actually be able to curate the files... (the client does not do this check). Expose `--assignee` end-to-end so you can also assign when using Github Actions UI.
- Enforce the latest contributor-friendly task naming convention.
- Case-insensitive `--template` matching for local templates in both scripts.
- Docs: added a "Wrapper vs. synapseclient" comparison table to `dev/DEVELOPMENT.md` spelling out what this wrapper adds beyond the client and why, and what it now delegates to the client.

#### Test plan
- [ ] Create a curation task via CLI with `create_curation_task.py`
- [ ] Create a curation task using GH Actions UI
- [ ] Create a recordset task via CLI with `create_recordset_task.py` (UI not offered for this one)

### DIFF
--- a/.github/workflows/create-curation-task.yml
+++ b/.github/workflows/create-curation-task.yml
@@ -128,8 +128,7 @@ jobs:
 
       - name: Install Synapse Python client
         run: |
-          # Install from develop branch which has curation task support
-          pip install git+https://github.com/Sage-Bionetworks/synapsePythonClient.git@develop
+          pip install 'synapseclient>=4.12.0'
 
       - name: Create curation task and bind schema
         id: create_task
--- a/.github/workflows/create-curation-task.yml
+++ b/.github/workflows/create-curation-task.yml
@@ -96,7 +96,7 @@ on:
         type: boolean
         default: false
       assignee:
-        description: 'Username, email, team name, or numeric principal ID to assign the task to (optional). Existing folder permissions are checked, not granted.'
+        description: 'Username, team name, or numeric principal ID to assign the task to (optional; email lookup is not supported by Synapse). Existing folder permissions are checked, not granted.'
         required: false
         type: string
         default: ''
--- a/.github/workflows/create-curation-task.yml
+++ b/.github/workflows/create-curation-task.yml
@@ -95,6 +95,11 @@ on:
         required: false
         type: boolean
         default: false
+      assignee:
+        description: 'Username, email, team name, or numeric principal ID to assign the task to (optional). Existing folder permissions are checked, not granted.'
+        required: false
+        type: string
+        default: ''
       create_issue:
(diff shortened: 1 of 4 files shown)


aditya-nath-sage (approved):

anngvu:
Thanks! 

1. `--assignee` can't resolve email because Synapse doesn't allow email lookups, so can't address that there
2. Will fix to add `pandas`

